PR #37, opened 2022-02-01
Fix colors of select component

pkretzschmar:


MohammadPCh (changes requested):
Thanks @pkretzschmar 
It's great.
I think it's better to fix borders in this pr too.
Now we have a blue and gray border together and should fix it based on figma
Screenshot from 2022-02-03 13-42-15
it's better to look like this:
Screenshot from 2022-02-03 13-45-31
Thanks

MohammadPCh (approved):
Thank you @pkretzschmar

PR #41, opened 2022-02-07
User public page

MohammadPCh:


src/apollo/types/types.ts
```diff
@@ -41,7 +41,20 @@ export interface IProjectCreation {
 }
 
 export interface IUser {
```
mateodaza, line 43: Some types here are number, no? Like totalDonated, totalReceived...

src/components/PaginationOld.tsx
```diff
@@ -0,0 +1,100 @@
+import { neutralColors, brandColors } from '@giveth/ui-design-system';
```
mateodaza, line 1: Should we keep the other pagination component if we are not using it?

mateodaza (changes requested):

src/components/PaginationOld.tsx
MohammadPCh: Thank you @mateodaza  for remembering this.

src/apollo/gql/gqlDonations.ts
```diff
@@ -22,6 +22,32 @@ export const FETCH_PROJECT_DONATIONS = gql`
...
+				id
+				firstName
+				lastName
+			}
+			project {
+				title
+			}
+			createdAt
+			currency
+		}
```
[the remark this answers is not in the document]
MohammadPCh: Oops. Thank you.

PR #48, opened 2022-02-13
Feat/reaction (project view and home page)

aminlatifi:


aminlatifi:
A project like/unlike is implemented in two views 
* Project view (by slug)
* Home page view

aminlatifi:
@MoeNick project update does not work! and projects page has a complex state management which needs refactoring in its project fetching part.

mateodaza (approved):
Awesome work man, thanks a lot.
Maybe for the future as I mentioned on the call we may use cookies instead of local storage to allow this call on the nextJs server side but that's not priority at the moment

mateodaza:
> @MoeNick project update does not work! and projects page has a convoluted state management which needs refactoring in its project fetching part.

Should we handle this on another issue? Will you take it or should I?

MoeNick:
> 

I think @aminlatifi is not going to continue it, shall we have another issue for it? I was assuming @pkretzschmar will take it on hearts issue https://github.com/Giveth/giveth-io-typescript/issues/74. I was wrong?

Issue #34, opened 2022-01-17
Toast Errors Components 


MoeNick:
# Design resources

Link to Figma GDS component-> [Toast component](https://www.figma.com/file/d3ciTS36toS9mra0RAPMe2/Giveth-Design-System-v0.1?node-id=1242%3A55067)

_You can find all possible use-case._

---

MoeNick:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/All-Devs-61644589b92edf0016f6999c/issues/Giveth/giveth-dapps-v2/34?planning-poker) with ZenHub @MohammadPCh

PR #54, opened 2022-02-14
Add Toasts

MohammadPCh:
- #34

### DIFF
--- a/src/components/toasts/index.tsx
+++ b/src/components/toasts/index.tsx
@@ -6,7 +6,7 @@ import {
 	semanticColors,
 	ButtonText,
 } from '@giveth/ui-design-system';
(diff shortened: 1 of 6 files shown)


MohammadPCh:
I add a test route for toasts.
you can check functionality there. and also we need to remove this route before merging.
thanks.
@mateodaza

mateodaza:
Loving that test route! Thanks for that

src/components/toasts/index.tsx
```diff
@@ -0,0 +1,179 @@
+import {
```
mateodaza, line 1: I'll approve because we need it asap but we should have this over the library right?

mateodaza (approved):

PR #64, opened 2022-02-18
Fix about page

pkretzschmar:
Ref Giveth/giveth-io-typescript#50

### DIFF
--- a/src/components/views/about/AboutHistory.tsx
+++ b/src/components/views/about/AboutHistory.tsx
@@ -22,11 +22,9 @@ const AboutHistory = () => {
 					by donations.
 				</Rect>
 				<UpperText>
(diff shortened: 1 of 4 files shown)


src/components/views/about/AboutHistory.tsx
```diff
@@ -1,62 +1,87 @@
...
+	place-items: center;
+	min-height: 170px;
+`;
+
+const ReadMoreButton = styled(ButtonLink)`
+	background-color: ${brandColors.pinky[500]};
+	margin-top: 32px;
+
+	&:hover {
+		background-color: ${brandColors.pinky[500]}80;
```
mateodaza, line 81: I think the 80 sneaked there

mateodaza (changes requested):

PR #68, opened 2022-02-23
Fix docs footer icon

pkretzschmar:
Ref Giveth/giveth-io-typescript#86﻿


MohammadPCh (approved):
Thank you @pkretzschmar

PR #116, opened 2022-03-01
implement givback toast new design

MohammadPCh:
- giveth/giveconomy#506

### DIFF
--- a/src/components/views/project/ProjectDonateCard.tsx
+++ b/src/components/views/project/ProjectDonateCard.tsx
@@ -13,16 +13,17 @@ import { Shadow } from '@/components/styled-components/Shadow';
 import CategoryBadge from '@/components/badges/CategoryBadge';
 import Routes from '@/lib/constants/Routes';
 import { mediaQueries, showToastError } from '@/lib/helpers';
-import QuestionBadge from '@/components/badges/QuestionBadge';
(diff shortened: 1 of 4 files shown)


src/components/views/project/ProjectDonateCard.tsx
```diff
@@ -255,11 +256,20 @@ const ProjectDonateCard = ({
...
 					<GivBackNotif>
-						<GLink size='Medium' color={brandColors.giv[300]}>
+						<Caption color={brandColors.giv[300]}>
 							When you donate to verified projects, you get GIV
 							back.
-						</GLink>
-						<QuestionBadge />
+						</Caption>
+						<a
+							href='https://docs.giveth.io/giveconomy/givbacks'
```
RamRamez, line 264: We may change the links in the future, or we may wanna use them in other pages. So I suggest add this url as "GIVBACK_DOCS" to the links file (lib/constants/links.ts).

RamRamez (changes requested):
Great, thanks.

src/components/views/project/ProjectDonateCard.tsx
MohammadPCh: great idea,
Thank you @RamRamez .

PR #176, opened 2022-03-13
check wallet on create project button

RamRamez:
check wallet on clicking create project button on header


MohammadPCh (approved):
Thank you @RamRamez

Issue #177, opened 2022-03-14
Improve Home Page SSR


aminlatifi:
The home page project list has a few issues because it's rendered server-side and the list of projects and their data are not updated! It affects
 - Total project reaction count. First, the old values are shown and then updated by new values because we fetch based on user data (we fetch once even if no user is available, e.g. no wallet is connected)
 - The order of projects is not updated. So when user loads the page, the projects reorder in short time after render.

aminlatifi:
This can be useful
https://www.redsunsoft.com/2021/10/simple-cache-expiration-for-apollo-graphql-queries/

divine-comedian:
@aminlatifi @MoeNick do we still need this for anything? Icebox? Keep on our sprints?

MoeNick:
I think we can compromise it. But lets wait for @aminlatifi 's comment

Issue #175, opened 2022-03-13
Homepage user like count changes after refresh


RamRamez:
On homepage, if user likes/unlikes a project, the like count changes after refreshing the page.

PR #178, opened 2022-03-14
Fix homepage totalReactions

aminlatifi:
Made totalReactions update on project card badge when props.totalReactions changes

It's a temporary solution for #175. It will not be needed if we solve #177


MohammadPCh (approved):
LGTM, Thank you @aminlatifi

PR #193, opened 2022-03-14
Add gnosis chain

divine-comedian:
replace xdai for gnosis chain on UI


MohammadPCh (changes requested):
Thank you @divine-comedian ,
It's great.
I think we should change it also in the Reward menu.

MohammadPCh (approved):
LGTM, Thank you @divine-comedian

Issue #192, opened 2022-03-14
Text fix on Join Page


laurenluz:
![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/dcfed90c-a215-4cad-89b7-ee3b5f2eafc5)

Just fix where the break is:

"Giveth is first and foremost a community of givers and changemakers. We are passionate people working together to build a crypto-economic system that can reward giving to good causes. Our project is open-source, decentralized, altruistic, and community-led.

Want to get more involved? Follow our social media and come say hello in a channel below, we look forward to welcoming you!"

laurenluz:
Should say "Join us on Discourse" (like the others)

![image](https://user-images.githubusercontent.com/75490971/158255673-9a932238-7b4f-4950-aa1b-279652854186.png)

laurenluz:
"Join us on Twitter"

![image](https://user-images.githubusercontent.com/75490971/158255776-dd087e8a-807a-4e6c-8a4c-dcaeecd2a0db.png)

laurenluz:
great, done

Issue #191, opened 2022-03-14
Support Giveth on About page button doesn't work


laurenluz:
![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/5300c312-d2f9-47ac-bfc3-c10b3c5397f2)

From [About](https://giveth-dapps-v2.vercel.app/about) page, this button should link to to the page to donate to Giveth... perhaps: https://giveth.io/donate/the-giveth-community-of-makers

Or is this not working because the Giveth project is not on v2?

laurenluz:
haha well, I hope this is working... testing on develop, the support link goes to: 
![image](https://user-images.githubusercontent.com/75490971/159953646-d686428d-45a1-4822-ae46-76b5ecf2342f.png)

but I assume on live it will go to the Giveth project donate page

right @pkretzschmar ?

Issue #180, opened 2022-03-14
UI issues


mosaeedi:
### Homepage issues

- [x] 1. "Tweet this" button -> Icon should be on the left side

![Screen Shot 2022-03-14 at 1 38 59 PM](https://user-images.githubusercontent.com/13708322/158155983-ec0f1718-cf0d-4be9-8363-4e29994de9c1.png)

- [x] 2. Project card shadow is missing

![Screen Shot 2022-03-14 at 1 39 52 PM](https://user-images.githubusercontent.com/13708322/158156110-bf782ed2-694b-4433-813a-a72abab62783.png)
![Screen Shot 2022-03-14 at 1 44 24 PM](https://user-images.githubusercontent.com/13708322/158156820-c2e8cb4d-fb5e-40b2-940d-2be960ace39b.png)
[Link to project component](https://www.figma.com/file/d3ciTS36toS9mra0RAPMe2/Giveth-Design-System-v0.1?node-id=2059%3A60081)

3. If I didn't connect my wallet yet when I tap on "create a project" I'll need to see the "Sign wallet" modal, instead I see the "complete profile" modal then by tapping ok I will be redirected to the onboarding screen and finally I receive the "Sign wallet" modal.
___

___

### Public profile

- [x] 2. Public profile misalignment 

![Screen Shot 2022-03-14 at 1 58 56 PM](https://user-images.githubusercontent.com/13708322/158159059-2dcb49dc-9b5f-4136-9a45-ed2832839529.png)

___

### My profile

- [x] 3. My profile misalignment

![Screen Shot 2022-03-14 at 1 59 57 PM](https://user-images.githubusercontent.com/13708322/158159208-52b8d20d-ae04-4cea-a89d-da6c8ed9a672.png)

___

PR #231, opened 2022-03-16
Fix minor UI issues

pkretzschmar:
Ref #180, #191, and #192


RamRamez (changes requested):
Great! Thanks @pkretzschmar 
I think in issue #180 the first item "Tweet this" is missed.

RamRamez:
I think responsiveness needs some improvements.
It's not related to this issue, but if you can fix them it would be great.
image

RamRamez (approved):

Issue #245, opened 2022-03-16
There are no tokens in the token list ??


laurenluz:
I'm connected to ropsten and testing on the [testing site](https://giveth-dapps-v2.vercel.app/donate/ymca-of-the-suncoast)

![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/c22f2ecf-f0c9-4862-b05f-b334aceb055f)

laurenluz:
This is a giving block project btw

mateodaza:
Are you on mainnet?

GriffGreen:
Ok so we need to tell them to switch to mainnet

mateodaza:
> Ok so we need to tell them to switch to mainnet

yes, working on a new modal now for that :)
Thanks for noticing, @MoeNick also told me that but we totally missed it for launch

GriffGreen:
Same issue
<img width="1105" alt="Screen Shot 2022-03-16 at 2 29 05 PM" src="https://user-images.githubusercontent.com/14945613/158674774-53c8e179-9e01-4ce0-b5bd-b44c46c704cc.png">

then when i switch to mainnet i get this: 
<img width="1256" alt="Screen Shot 2022-03-16 at 2 31 09 PM" src="https://user-images.githubusercontent.com/14945613/158674900-20395cd6-accd-4aa0-976f-06c44e9f7fd2.png">

GriffGreen:
Then when i close the modal i still see no tokens... even after refresh
<img width="1385" alt="Screen Shot 2022-03-16 at 2 32 03 PM" src="https://user-images.githubusercontent.com/14945613/158675056-bd816a69-b3c2-4905-81d4-5052456788ba.png">

mateodaza:
On it!

MoeNick:
I'm still having problems with donations to these giving block projects
the token list on both chains are empty. 
the modal also appears when I'm on the right network. 
@mateodaza @pkretzschmar

mateodaza:
This is not on staging yet

mohammadranjbarz:
@mateodaza  You can get valid tokens for project from backend
https://github.com/Giveth/impact-graph/issues/380#issuecomment-1068337179

divine-comedian:
tokens show up on ropsten now 

![image](https://user-images.githubusercontent.com/67759413/161101659-0f97a2f7-ae9e-48b3-9227-c8c3d8b4dd93.png)

[...]

PR #311, opened 2022-03-21
H 245 no tokens in the token list

RamRamez:
Ref. #245


mateodaza:
Thanks a lot @RamRamez ! A couple of things I noticed 
- Selected token icon won't show up on item
 
image

- Duplicated GIV token on list is back, it was previously fixed
image

- For giving block projects I can't see the list on mainnet, is it because it's staging?

RamRamez:
I solved the first and second one. The last will be ok on production, since we fetch token list from backend and backend doesn't have any tokens for Ropsten for giving block projects

mateodaza (approved):

Issue #244, opened 2022-03-16
Connected to Ropsten (testing on v2), tells me to switch to gnosis chain for a giving block project


laurenluz:
this popup makes no sense, Giving block projects don't even accept funds on gnosis chain.

![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/9ded9ee7-69d6-43d4-bd3f-cc90a327406e)

pkretzschmar:
@laurenluz I changed the modal behavior. Now, if you go to a project that belongs to GivingBlock, and you're not connected to Ethereum, will ask you to change to ETH chain, and if you are trying to donate to another project that is not from Giving Block, will suggest to use Gnosis or Ethereum.

divine-comedian:
I just made a small update to the text - other than that the changes are working as expected! 

on regular projects we get this modal 
![image](https://user-images.githubusercontent.com/67759413/161094369-71feae1f-f242-4f17-b58e-d3f09fa7d5ea.png)

on giving block projects we receive a prompt just for ethereum mainnet

![image](https://user-images.githubusercontent.com/67759413/161094682-75d34f49-be3b-40e5-b368-7187f648cc2c.png)

resolved! moving to done

PR #359, opened 2022-03-25
Fix suggested networks on modal

pkretzschmar:
Ref #244﻿

### DIFF
--- a/src/components/modals/ChangeNetwork.tsx
+++ b/src/components/modals/ChangeNetwork.tsx
@@ -30,7 +30,7 @@ export const ChangeNetworkModal: FC<IChangeNetworkModalProps> = ({
 	const NetworkName =
 		targetNetwork === config.MAINNET_NETWORK_NUMBER
 			? 'Ethereum'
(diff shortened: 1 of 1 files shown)


src/components/modals/ChangeNetwork.tsx
```diff
@@ -30,15 +30,18 @@ export const ChangeNetworkModal: FC<IChangeNetworkModalProps> = ({
 	const NetworkName =
 		targetNetwork === config.MAINNET_NETWORK_NUMBER
 			? 'Ethereum'
-			: 'Gnosis Chain';
+			: 'Ethereum or Gnosis Chain';
```
mateodaza, line 33: Remember that this modal will also be used on The Giving Block projects. If we show this message it won't make much sense. Can we make this modal receptive so for this case it shows Ethereum and for the others "Ethereum or Gnosis Chain"

mateodaza (changes requested):

mateodaza (approved):

PR #370, opened 2022-03-28
Sentry config

aminlatifi:
Read sentry setting from environmental variables

### DIFF
--- a/src/utils/Logger.ts
+++ b/src/utils/Logger.ts
@@ -1,10 +1,9 @@
 import { init, captureException } from '@sentry/browser';
 
 if (typeof window === 'object') {
-	// const sentryId = process.env.SENTRY_ID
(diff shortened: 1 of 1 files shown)


src/utils/Logger.ts
```diff
@@ -1,10 +1,9 @@
 import { init, captureException } from '@sentry/browser';
 
 if (typeof window === 'object') {
-	// const sentryId = process.env.SENTRY_ID
-	const sentryId = 'de31cb89ac0045afbb9d28322cc9c040@o510515';
+	const sentryDsn = process.env.SENTRY_DSN;
```
mateodaza, line 4: Are you sure this variable can be accessed from the client side? Next requires the prefix NEXT_PUBLIC_ so it's available, if we leave it like this I think the value will return empty

mateodaza (changes requested):

mateodaza (approved):

Issue #399, opened 2022-03-31
If project is a a draft, listing status should say "draft"


laurenluz:
Related to #287... I imagine listing=null for draft projects, but the status "waiting for review" here makes no sense. If it's a draft, it should say "Draft"

![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/5217aa1f-1659-4f06-a399-3162bb3f04dd)

MoeNick:
I think the business is miss-understood, waiting for review should still be waiting for review. 
Draft is another concept. 

We gonna change the design for better understanding: 
![image.png](https://images.zenhubusercontent.com/604f242c2819d5da6426ad03/36f53297-579-405e-b3fe-6f94dd3d8ff1)

divine-comedian:
testing on `staging.giveth.io`

I can clearly distinguish on the `my projects` page the listing status of my projects - 
`draft` shows up separately from a published project that is still `waiting for review`

![image](https://user-images.githubusercontent.com/67759413/163680603-97fe22c3-6217-45e8-a6b3-c5b942226b5c.png)

are you still going to change the design @MoeNick  or is this issue complete?

MoeNick:
It should not be like this. Im sorry that it merged in this way. Is there any chance we revert it back? @pkretzschmar

divine-comedian:
> It should not be like this. Im sorry that it merged in this way. Is there any chance we revert it back? @pkretzschmar

should this issue still be tested or moved to backlog?

MoeNick:
I asked @RamRamez to implement the new design here: https://github.com/Giveth/giveth-dapps-v2/pull/527/files/c2abfab54145a994135c43b743dd34cde371d79b..2999ab9f2c5025ad34d389f11cc772e5de7c9bd1

divine-comedian:
nice! new design looks good - just pushed one spelling change  `deactive` -> `deactivated`
![image](https://user-images.githubusercontent.com/67759413/163813758-384d5440-345a-4fe8-ab40-cb4eacb09e0f.png)

i think we can call this done!

PR #403, opened 2022-04-01
Fix badge label

pkretzschmar:
Ref #399﻿


src/components/views/userPublicProfile/ProjectsTable.tsx
```diff
@@ -66,7 +66,7 @@ const ProjectsTable: FC<IProjectsTable> = ({
 			title = 'Listed';
 		} else if (listed === null) {
 			color = semanticColors.blueSky;
-			title = 'Waiting for review';
+			title = 'Draft';
```
RamRamez, line 69: What if a project is published but not listed yet? Do you show `Draft` for that?
Maybe you should add another condition to check if the project is `drafted` or not

RamRamez (changes requested):

src/components/views/userPublicProfile/ProjectsTable.tsx
pkretzschmar: Added the draft case and maintained the Waiting for review tag.

RamRamez (approved):

Issue #385, opened 2022-03-30
Review homepage content allignment


jahanzaib-jay:
On the home page, under the GIVeconomy section, the alignment of subheadings "For projects" and "For Givers" seems a bit off. 

"For Givers" part should be slightly moved further right- similar to the "Calling all changemakers section.

![Screenshot_4.png](https://images.zenhubusercontent.com/620d7be8f7acb77200f9d2d9/a69ff744-8084-4f6c-b15d-6040485987ec)

![Screenshot_5.png](https://images.zenhubusercontent.com/620d7be8f7acb77200f9d2d9/4f76bc06-96f8-402c-a97c-080174e88a7f)

jahanzaib-jay:
@mosaeedi Can you please have a look?

divine-comedian:
the alignment will vary depending on your screen resolution but everything seems to look good! 
![image](https://user-images.githubusercontent.com/67759413/163680214-39f70223-e2e0-4063-b001-e7bb73ce8dc5.png)

moving to DONE

PR #405, opened 2022-04-01
Fix homepage allignment

pkretzschmar:
Ref #385﻿

### DIFF
--- a/src/components/views/homepage/HomeChangeMakers.tsx
+++ b/src/components/views/homepage/HomeChangeMakers.tsx
@@ -64,12 +64,14 @@ const HomeChangeMakers = () => {
 			<PurpleArc />
 			<CyanArc />
 			<MustardArc />
-			<D3>Calling all Changemakers!</D3>
-			<br />
-			<Lead>
-				Do you have a for-good project that&apos;s creating value for
-				society, for the environment, or for the world?
-			</Lead>
+			<TitleContainer>
+				<D3>Calling all Changemakers!</D3>
+				<br />
+				<Lead>
+					Do you have a for-good project that&apos;s creating value
+					for society, for the environment, or for the world?
+				</Lead>
+			</TitleContainer>
 			<MiddleSection>
 				<Lead>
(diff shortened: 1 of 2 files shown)


src/components/views/homepage/HomePurpleSection.tsx
```diff
@@ -187,6 +187,7 @@ const ArcSmall = styled(Arc)`
 const ForMakersGivers = styled(Row)`
 	margin-top: 107px;
 	margin-bottom: 50px;
+	justify-content: space-between;
```
RamRamez, line 190: there is too much space between them. maybe it's a better idea too set a gap between the two sections instead of `space-between`

image

src/components/views/homepage/HomeChangeMakers.tsx
```diff
@@ -64,12 +64,14 @@ const HomeChangeMakers = () => {
 			<PurpleArc />
 			<CyanArc />
 			<MustardArc />
-			<D3>Calling all Changemakers!</D3>
-			<br />
-			<Lead>
-				Do you have a for-good project that&apos;s creating value for
-				society, for the environment, or for the world?
-			</Lead>
+			<TitleContainer>
```
RamRamez, line 67: why did you add this?

RamRamez (changes requested):

src/components/views/homepage/HomeChangeMakers.tsx
pkretzschmar: Sorry, I forgot to remove, I was just testing some stuff.

src/components/views/homepage/HomePurpleSection.tsx
```diff
@@ -200,7 +201,7 @@ const ForMakersContainers = styled(Col)`
 
 const ForMakersButton = styled(Button)`
 	width: 300px;
-	margin-top: 24px;
+	margin-top: auto;
```
RamRamez: It's not true according to figma:

image

RamRamez (approved):

PR #421, opened 2022-04-04
Making some improvements for performance

mateodaza:
_app.js right now is the heaviest load on the app, we need to make it lighter some how

issue related > #422 

hopefully with this changes we can get this values on production (this is on a local build running lighthouse)

image

Also maybe we should improve the SEO with another issue itself


MohammadPCh (approved):
Thank you @mateodaza

Issue #419, opened 2022-04-02
RegenFarms Transaction Status modal displays 'Add GIV to MetaMask' but should display 'Add [Reward Token] to MetaMask'


willyogo:
<img width="845" alt="image" src="https://user-images.githubusercontent.com/26232238/161396387-478d055c-d561-49a0-825e-2ff42eddcb3e.png">

AC:
1. The transaction status modal for any regen projects should have a button to 'Add [Reward Token] to MetaMask' with Reward Token matching the corresponding RegenFarm

laurenluz:
@MoeNick would be good to get this one fixed asap

divine-comedian:
testing on `staging.giveth.io` 

the modal appears to be update on tx pending and tx successful screen
![image](https://user-images.githubusercontent.com/67759413/163844207-3ad1d286-1b49-41b2-a551-9917e848f487.png)

moving to DONE

PR #424, opened 2022-04-05
Supported RegenFarm add token

aminlatifi:
Changed AddGIVToken component to a more generalized AddToken which support others token too and not only GIV
Refs https://github.com/Giveth/giveth-dapps-v2/issues/419


MohammadPCh (approved):
LGTM, Thank you @aminlatifi

Issue #417, opened 2022-04-04
Can't donate with Wallet Connect


MoeNick:

aminlatifi:
I have 1 GIV in my multisig wallet, but donate view doesn't recognize it
<img width="1433" alt="image" src="https://user-images.githubusercontent.com/5684607/163791207-9cda3b64-64c8-4418-837e-f607756b5563.png">

<img width="916" alt="image" src="https://user-images.githubusercontent.com/5684607/163791267-b04d434c-c139-4901-a7d0-0f7cbe019186.png">

aminlatifi:
If the grpc node url is the issue, we can use a public one like this
https://rpc.ankr.com/eth

divine-comedian:
testing on `staging.giveth.io` 

appears to be having the same issue as before. I can successfully connect my gnosis-safe with walletConnect

we can see on staging it shows my DRGIV3 balance in the top corner, however in the donate modal it is not detecting any of my tokens, making it impossible to donate.

![image](https://user-images.githubusercontent.com/67759413/169868088-6f81c892-430d-4342-a265-dc76bc74f2a2.png)

moving back to backlog

divine-comedian:
@MoeNick 

Do we have any current Wallet Connect integration? This is used often by mobile apps as well, not only multisigs

MoeNick:
It was working but we removed it due to the bug that I remember it was for the grpc node url and should use public one.
https://rpc.ankr.com/eth

However, we decided to do it again here as you brought it up: https://github.com/Giveth/giveth-dapps-v2/issues/3053

MoeNick:
we can close this one

Issue #333, opened 2022-02-28
Add approve+stake option to givfarm staking


aminlatifi:
As a user, I can choose to stake on unipool (givfarm pools) by two transactions of approve+stake instead of a single transaction of `stakeWithPermit`.
It is useful when a multisig user wants to stake and he/she cannot sign a message in advance.

Refs https://github.com/Giveth/GIVeconomy/issues/502

mosaeedi:
### Design resource

[Modal with infinity approval](https://www.figma.com/file/d3ciTS36toS9mra0RAPMe2/Giveth-Design-System-v0.1?node-id=488%3A47168)

pkretzschmar:
@mosaeedi @aminlatifi this toggle shouldn't refer to the stake with a permit or without a permit?

aminlatifi:
> @mosaeedi @aminlatifi this toggle shouldn't refer to the stake with a permit or without a permit?

You are right @pkretzschmar. @mosaeedi this design is different than the topic of this issue!

PR #431, opened 2022-04-05
Feat approve and stake

pkretzschmar:
Ref #333, and #417


aminlatifi (changes requested):
I think two independent issues are solved in this single PR.
1. approve/permit switch button
2. gnosis-safe support

They can be addressed in different PRs 
Also, for fixing gnosis-safe issue we have added dependencies which makes our bundle bigger.

aminlatifi:
@pkretzschmar Can you create a separate PR just for adding approve/permit mode switch feature?

Issue #428, opened 2022-04-05
GIVfarm total rewards are not being calculated properly


laurenluz:
It seems like it's excluding GIV / HNY on gnosis chain

A user reported it:
![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/a35d5e1c-0f05-4bf7-acf4-9fd205832988)

And it's happening for me as well:
![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/015970cf-1fc2-4bb2-9b0d-7b938224780f)

@aminlatifi can you have a look?

MoeNick:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/All-Devs-61644589b92edf0016f6999c/issues/Giveth/giveth-dapps-v2/428?planning-poker) with ZenHub @aminlatifi @MohammadPCh

aminlatifi:
@MoeNick Made a PR which solves it.

laurenluz:
looking good on staging

![image](https://user-images.githubusercontent.com/75490971/165184271-fd93a092-9944-4052-a589-6bf29996aee4.png)

PR #432, opened 2022-04-06
Only non regen farms reward are included in farm total GIV reward

aminlatifi:
Fixed linting warnings
Moved FarmProvider to FarmView page
Refs #428


aminlatifi:
@mateodaza It's a high priority fix and simple to solve, so made to main branch

PR #433, opened 2022-04-06
Only non regen farm rewards are included in regen farm total GIV reward

aminlatifi:
Moved Farm Provider to FarmView
Refs #428


aminlatifi:
It's a high priority issue, so I made the update to main first and then will apply on develop

mateodaza (approved):

Issue #189, opened 2022-03-14
GIVeconomy: Adding steps to the Stake modal


mosaeedi:
We need to add the steps to the stake modal to let users know the staking happens in two steps.

Link to the Figma component

[https://www.figma.com/file/d3ciTS36toS9mra0RAPMe2/Giveth-Design-System-v0.1?node-id=488%3A47225](https://www.figma.com/file/d3ciTS36toS9mra0RAPMe2/Giveth-Design-System-v0.1?node-id=488%3A47225)

aminlatifi:
@mosaeedi Would you please add requirements of https://github.com/Giveth/liquidity-mining-dapp/issues/188 to that design too?

divine-comedian:
very slick update - the modal seems to working according to the figma - 
![image](https://user-images.githubusercontent.com/67759413/163639098-4962a923-a629-44f7-aaa2-89984caae5ae.png)

call this DONE!

PR #442, opened 2022-04-10
Adding steps to stake modal

MohammadPCh:
- #189


aminlatifi (changes requested):
When the user uses a percentage button, its color should be changed
image

src/components/modals/Stake.tsx
```diff
@@ -173,6 +176,47 @@ export const StakeModal: FC<IStakeModalProps> = ({
...
+															StakeState.WRAPPING
+													)
+												}
+											>
+												2
+											</StakeStepNumber>
+										</StakeStep>
+									</StakeStepsContainer>
+								) : (
+									<StakeStepsPlaceholder />
```
aminlatifi, line 218: Do we need this?

MohammadPCh: Yes because of Design we need space there.

aminlatifi (approved):

PR #501, opened 2022-04-12
Max Button and Percentage Button integration

aminlatifi:
Max button makes the 100% button high light


MohammadPCh (approved):
Thank you @aminlatifi

PR #511, opened 2022-04-13
Fix change network conditions

RamRamez:


mateodaza (commented):
image

I'm getting this weird message when I'm on a different network, should it say both ropsten and gnosis? when I click change network it sends me to ropsten

RamRamez:
> image
> 
> @RamRamez I'm getting this weird message when I'm on a different network, should it say both ropsten and gnosis? when I click change network it sends me to ropsten

fixed

PR #524, opened 2022-04-17
Fixed Toast size on mobile devices

ascode94:
Now it looks like this on Mobile devices:

image


MohammadPCh (approved):
Thank you @ascode94

Issue #542, opened 2022-04-19
Connect to previous selected wallet automaticly


MohammadPCh:
go through these steps:
1. connect your Metamask wallet
2. disconnect from it.
3. connect your Torus wallet.
4. exit the app
now the expected behavior is when you go to the site you connect automatically to Torus Wallet but first, you connect to meta mask the disconnect after that site asks you to connect your Torus Wallet.

PR #543, opened 2022-04-19
move eager connet to userProvider

MohammadPCh:
- #542


src/context/UserProvider.tsx
```diff
@@ -51,15 +51,24 @@ const UserContext = createContext<IUserContext>({
...
 
+	useEffect(() => {
+		const selectedWalletName = localStorage.getItem(StorageLabel.WALLET);
+		const wallet = walletsArray.find(w => w.value === selectedWalletName);
+		if (wallet) {
+			activate(wallet.connector, err => {
+				console.log('err', err);
+			});
+		}
+	}, [activate]);
```
RamRamez, line 70: `activate` is a always a function and I think adding it as a dependency is not necessary. If user locks his wallet and unlock again, the app won't try to `activate`, So I think you should replace `activate` dependency with web3React `active`

RamRamez (changes requested):

src/context/UserProvider.tsx
```diff
@@ -13,9 +13,9 @@ import { initializeApollo } from '@/apollo/apolloClient';
 import { GET_USER_BY_ADDRESS } from '@/apollo/gql/gqlUser';
 import { compareAddresses, showToastError, signMessage } from '@/lib/helpers';
 import { getToken } from '@/services/token';
-import useWallet from '@/hooks/walletHooks';
```
RamRamez, line 16: this file is not used anymore, so I think you can remove it completely.

RamRamez (changes requested):

src/context/UserProvider.tsx
MohammadPCh: exactly ;)
Thank you @RamRamez

src/context/UserProvider.tsx
MohammadPCh: I put this for the time that the activate function changed, for example, if it is not initial yet.

RamRamez (approved):

PR #549, opened 2022-04-19
Add approve/stake mode

pkretzschmar:
Ref #333


mateodaza (commented):
It's all working properly! Great work, I have a question
Is it only working for Kovan / Mainnet ? The toggle won't appear for xDAI to me

src/hooks/walletHooks.ts
```diff
@@ -0,0 +1,92 @@
+import { useState, useEffect } from 'react';
```
RamRamez, line 1: We had deleted the `walletHooks.ts` component completely. Did you update your branch?
Do you wanna restore this component? since it's not used anywhere.

RamRamez (changes requested):

src/hooks/walletHooks.ts
RamRamez: @mateodaza JFYI

pkretzschmar: I was looking at it, and I don't need to. I removed the file.

PR #551, opened 2022-04-19
Welcome modal shown only when user actually logs in

mateodaza:
Modal was being shown even when rejecting the login


MohammadPCh (changes requested):
Thank you @mateodaza ,
I think it's better to change this line instead of changing FirstWelcomeModal.
WalletModal.tsx line: 42
```
activate(selected.connector, showToastError, true)
```

mateodaza:
> Thank you @mateodaza , I think it's better to change this line instead of changing FirstWelcomeModal. WalletModal.tsx line: 42
> 
> ```
> activate(selected.connector, showToastError, true)
> ```

Thank you! Will do

PR #582, opened 2022-04-21
Set startTime for giv/dai farm

aminlatifi:
a little refactoring


MohammadPCh (approved):
Nice, Thank you @aminlatifi

PR #602, opened 2022-04-24
Reset GivEconomy video after ended and added separate file for stream…

ascode94:
Refactored video to use react useRef instead of document.getElementById and disabled draggable and user-select of the Video Play button image.
Created separate file for givStream Cards and use it in Overview section


MohammadPCh (approved):
LGTM, Thank you @ascode94

Issue #397, opened 2022-03-31
Support period finish time in calculating unipool reward and APR


aminlatifi:
Unipool reward rate becomes zero when the period finishes and the unipool is not notified of a new reward to pay.
It's should be supported in UI, because the APR particularly doesn't become zero!

PR #603, opened 2022-04-24
Support periodFinish in UnipoolHelper reward

aminlatifi:
UnipoolHelper returns zero reward amount after the periodFinish timestamp
Refs #397


MohammadPCh (approved):
Nice, Thank you @aminlatifi

PR #632, opened 2022-04-28
do not show similar projects when there's nothing

RamRamez:


mateodaza (approved):
Working fine, thank you

PR #637, opened 2022-04-28
Some changes on donate component

RamRamez:
Ref. #641

1. require sign for anonymous users also
2. remove unused saveDonationTransaction query
3. show error to user if donation is not saved
4. remove setInProgress and setUnconfirmed and wait till tx is mined then show success/error
5. complete refactor send transaction related functions
6. capture if tx is cancelled, failed, mined or speed up
7. remove requesting to network every one second for tx state and add listen instead
8. remove unused codes


MohammadPCh (approved):
LGTM, thank you @RamRamez

Issue #646, opened 2022-05-01
Add more linters


MohammadPCh:
We need more linter to improve our code quality.
I suggest having a linter for sort of import and unused imports.

PR #647, opened 2022-05-01
Add more linters

MohammadPCh:
- #646


ascode94 (approved):
Thanks @MohammadPCh For adding eslint plugins and rules

PR #649, opened 2022-05-01
Added captureException to all catch blocks for having them in sentry

ascode94:
Handled all catch blocks for sentry


MohammadPCh (approved):
Grea Job, Thank you @ascode94

PR #661, opened 2022-05-04
F/ Displays archived pools on toggle + unescaping characters on project cards

mateodaza:


MohammadPCh (changes requested):
Thank you @mateodaza 
The Code quality is great. but I have 2 UX issues.

1. I think we should show archived pools after the GIV pool.
2. I think we should have a switch button instead of a radio button.

I mention @markoprljic and @laurenluz to know their opinion.

mateodaza:
> Yeah! I agree with #1, it would be good to show archives pools last.
> 
> As for #2, yeah I think the swtich button would be nicer, but I don't feel strongly. Maybe @markoprljic has an opinion.

I'll make these changes :)

MohammadPCh:
@markoprljic 
This is what we have now:
Screenshot from 2022-05-05 20-15-38
Screenshot from 2022-05-05 20-15-45
and I suggest something like this:
image

markoprljic:
Toggle button would be more appropriate in this context. 
Active pools <> Archived pools

But, since we are using the toggle button for chain selection, I'd say we go with the Switch. @mosaeedi need to create it in GDS if not already available. 

If we plan to support more chains in the future we should consider changing the toggle button for chains to a dropdown.
When we do that then the toggle for archived pools would be better option.

Hope that makes sense.

mateodaza:
> Switch

Ok! @mosaeedi let me know, I think it's not available on the DS yet, thank you

MohammadPCh:
Ok @mateodaza, Do you agree to fix number 1 (move archived pools to end) in this PR and postpone implementing the switch button until later?
because I worry about conflicts.

PR #665, opened 2022-05-04
remove unused images

RamRamez:


src/components/Header/Header.tsx
```diff
@@ -141,172 +141,166 @@ const Header: FC<IHeader> = () => {
...
-									<a>
-										<Logo>
-											<Image
-												width='48px'
-												height='48px'
-												alt='Giveth logo'
-												src={`/images/logo/logo1.png`}
-											/>
-										</Logo>
-									</a>
```
MohammadPCh, line 176: we need the anchor tag because it's adding another internal link for us.so I think it's better to keep it.
@RamRamez

MohammadPCh (changes requested):

MohammadPCh (approved):
Thank you @RamRamez

Issue #686, opened 2022-05-12
Eslint rule for imports


alireza-sharifpour:
Problem: Currently we are not getting eslint error if we import something that doesn’t exist. We are just getting typescript error for that.
Example:
If I remove this function, the only error I get is:

![Screen Shot 2022-05-12 at 1 11 47 PM](https://user-images.githubusercontent.com/48887817/168029691-6277c5a5-a68a-4f93-a97a-dd49aef980a6.png)

So when we run linter we cannot find out the problem.

PR #687, opened 2022-05-12
Added new eslint rule for named-imports

alireza-sharifpour:
#686


MohammadPCh (approved):
Thank you @alireza-sharifpour

Issue #693, opened 2022-05-14
notify backend about donation status


RamRamez:
Remove saveDonation and add createDonation endpoint.
Notify backend on status change
Add donation status to tables

Ref [#507](https://github.com/Giveth/impact-graph/pull/507)

divine-comedian:
how can this issue be tested?

RamRamez:
> how can this issue be tested?

When you donate and tx is NOT mined yet, your donation status (in My account or in donation tab in project view) should be `pending`. If tx is mined and the app window in browser in still open, the App will show you the tx success view then notify the backend about tx status update, then tx status in donation tables should be `verified`. If you close the window before tx being mined, App can not notify backend and backend will check for status update every one hour. So in this case your donation status will be updated after one hour.
This situation should be similar for failed tx (like cancelled tx) or speed up tx.

we will remove donation status column in future.

divine-comedian:
testing on `staging.giveth.io` 

I was able to make a few donations using low gas fees and could confirm that while the tx was still being mined on  the my donation screen it would show as `pending` - once the transaction was mined it would change to `verified` 

I could also confirm that while the tx was still pending no call was made to segment. when the tx was confirmed then the call was sent to segment, triggering the email. this flow seems to be work I would only suggest one semantic change

would it be possible to change the terminology of `verified` to `confirmed`? We should avoid confusing users between "verified projects" and "verified transactions"

RamRamez:
> 

I can change it in front-end but I think it's not the best choice.
@MoeNick @mohammadranjbarz what do you think?

mohammadranjbarz:
> testing on `staging.giveth.io`
> 
> I was able to make a few donations using low gas fees and could confirm that while the tx was still being mined on the my donation screen it would show as `pending` - once the transaction was mined it would change to `verified`
> 
> I could also confirm that while the tx was still pending no call was made to segment. when the tx was confirmed then the call was sent to segment, triggering the email. this flow seems to be work I would only suggest one semantic change
> 
> would it be possible to change the terminology of `verified` to `confirmed`? We should avoid confusing users between "verified projects" and "verified transactions"

1. So if it didn't change and still is **pending** it's a bug, please put your donation info or your txHash here to investigate more about that
2. Seems fine I can change it to `confirmed` ,  @MoeNick Is it ok?

divine-comedian:
> > testing on `staging.giveth.io`
[...]

PR #694, opened 2022-05-14
new service for donation save and donation update

RamRamez:
Ref. #693


MohammadPCh (approved):
Nice Job, Thank you @RamRamez

Issue #733, opened 2022-05-17
Cannot find module './***.png'


MohammadPCh:
Sentry Link:
https://sentry.io/organizations/giveth-ds/issues/3261985762/?project=6292414
![Screenshot from 2022-05-17 17-22-33](https://user-images.githubusercontent.com/17097576/168815254-a7786e50-0a7d-4f4a-98c3-e9ac3c3bac76.png)

PR #736, opened 2022-05-17
remove captureException from token picker icon catch function

RamRamez:
Ref #733


alireza-sharifpour (approved):
Thanks Ramin 😊 @RamRamez

PR #768, opened 2022-05-19
Integrates authentication microservice using SIWE signatures

mateodaza:


src/components/menu/MenuWallet.tsx
```diff
@@ -137,7 +138,10 @@ const MenuWallet = () => {
 					))}
 					{isSignedIn && (
 						<MenuItem
-							onClick={() => dispatch(signOut())}
+							onClick={async () => {
+								await logoutToken(token!);
```
MohammadPCh, line 142: I think we should move this to the signOut reducer

src/services/token.ts
```diff
@@ -44,6 +44,27 @@ export async function getToken(
...
+		},
+	);
+}
+
+export async function logoutToken(jwt: string) {
+	return await postData(`${config.MICROSERVICES.authentication}/logout`, {
+		jwt,
+	});
+}
+
```
MohammadPCh, line 67: I think we can move these to user feature folder

MohammadPCh (changes requested):
It's a great job, I added 2 suggestions also I will create a PR for this branch soon.

MohammadPCh (approved):
Great Job, Thank you @mateodaza

Issue #774, opened 2022-05-22
set a valid amount error in donation page


RamRamez:
It says "set a valid amount" but still I can donate

<img width="344" alt="image" src="https://user-images.githubusercontent.com/73441857/169698216-1ff53c73-f9e1-47fc-90e4-9dd3869e4976.png">

divine-comedian:
testing on `staging.giveth.io` 

now on the donate page when I have an invalid amount, as in too many decimal places the donate button is no longer clickable and I cannot donate.

![image](https://user-images.githubusercontent.com/67759413/169842103-4b49d7d3-c214-4b4a-889b-d385cd76239e.png)

seems to be resolved! moving to DONE

PR #775, opened 2022-05-22
Fix/774 set valid amount for donate

RamRamez:
Ref. #774


alireza-sharifpour:
@RamRamez Please just resolve the conflict.

Issue #557, opened 2022-04-20
Unfinished project - no draft saved


missgene:
I tried creating a project entitled "Quotes of the day" and I only added two photos to the description. I intentionally closed the tab and all the entries I made were not saved. 

I went to My Projects under my account page and I did not see that a project is pending or in the draft. 

I think this is important as some users may accidentally close their browser while creating a project. Also, I noticed that there was no dialog box that asked me something like "Are you sure you want to close this window?" or "You have an unfinished project. Closing this window will delete your progress", etc. 

@MoeNick

MoeNick:
@MohammadPCh @RamRamez @mateodaza 
Any idea of autosaving the project right after the user moves from any field to another? 
I need your UX opinions rather than technical ones.

RamRamez:
@MoeNick 
Like what we had in next.giveth.io we can store changes to browser storage.
If project is created or drafted, the local storage will be deleted and empty form will be loaded on create project, otherwise local storage will be filled in form.

MohammadPCh:
@MoeNick 
I think the suggestion is great. we can ask the user before closing the tab.

MohammadPCh:
@MoeNick  I create something like this:
![Screenshot from 2022-05-02 19-00-49](https://user-images.githubusercontent.com/17097576/166252073-49134b90-835d-4934-979f-9230d768f451.png)
FYI we can't change anything here. but I think this is general UX that users know it.

MoeNick:
@Cherik#4711 jan, did you push the changes? I could not test it on the staging, or production

MohammadPCh:
@MoeNick Jan, No, I didn't. I say it's possible, but we should consider changing the route. do we need a similar modal?

MoeNick:
@MohammadPCh can I move it to UAT?

MohammadPCh:
@MoeNick No, I'm busy with other issues. Let me tell you when it is finished.

divine-comedian:
testing on `staging.giveth.io` 
brave browser

I have opened a project and begin filling information, I go to close the tab or the browser window and I receive the dialogue asking me if im sure

![image](https://user-images.githubusercontent.com/67759413/172267338-482685a7-167e-4f93-8542-06edfcffb003.png)

to confirm @MohammadPCh and @MoeNick there was no plans inside of this issue to save state and users would return to their unfinished draft after closing the window? it was unclear from the comments

PR #777, opened 2022-05-22
Ask before close on create project

MohammadPCh:
- #557


mateodaza (approved):
awesome work, thanks a lot

Issue #738, opened 2022-05-17
GIVpower - Explanatory Page


laurenluz:
This page is ready to be built! It's just an explainer page.

We still don't have all the links set up, or the rewards for the harvest section, but perhaps we can start implementing the design?

![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/2e414a31-d9a0-44a8-ab68-df38455441f3)

[Link to figma](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=8207%3A169976)

@MoeNick @mateodaza

MoeNick:
Is it final Lauren, I can see it with Review tab in Figma

laurenluz:
@alireza-sharifpour on the GIVpower page, the rewards box needs to be updated...

If the user has their wallet connected & on the wrong chain, it should say:
"GIVpower is only available on Gnosis Chain. Please switch the network."
"Switch to Gnosis Chain"

If the user does not have their wallet connect, it should say:
"GIVpower is available on Gnosis Chain. Please connect your wallet."
"Connect Wallet"

laurenluz:
Also, this should say "BROWSE PROJECTS"

![image](https://user-images.githubusercontent.com/75490971/171269315-2bd44104-42d0-4b02-adf1-1202e1cb68f3.png)

alireza-sharifpour:
Thanks, @laurenluz, for checking this page.
I will fix the issues you mentioned.
Also, I have a question about the texts at the bottom. Are there links somewhere? if they are, where should they link to? [GET GIVPOWER, BOOST PROJECTS, SEE REWARDS, BROWSE PROJECTS]
<img width="1121" alt="Screen Shot 2022-06-01 at 10 42 08 AM" src="https://user-images.githubusercontent.com/48887817/171339257-1395122d-7737-4d09-aa75-69da867fcdf6.png">
<img width="1280" alt="Screen Shot 2022-06-01 at 10 44 57 AM" src="https://user-images.githubusercontent.com/48887817/171339648-eeeb0309-4dc0-41be-b040-045159c50573.png">

divine-comedian:
where can this issue be tested? is there a separate deployment?

alireza-sharifpour:
> where can this issue be tested? is there a separate deployment?

This issue is just UI related stuff, the upper card and modal is in issue #763 and after that we can connect the upper card to staking modal.
[...]

PR #780, opened 2022-05-23
Added Givpower TOP and base files

alireza-sharifpour:
#738


MohammadPCh (approved):
LGTM, Thank tou @alireza-sharifpour

PR #800, opened 2022-05-25
Feat/752 project registry

alireza-sharifpour:
#752


RamRamez (changes requested):
Great, thanks @alireza-sharifpour

RamRamez:
focus border color should be giv[500]
I think it's better to set the width of drop-down menu equal to select component
also please check the drop-down menu hover color with the design system

image

RamRamez:
Please also update the border color of the description input component

alireza-sharifpour:
Thanks @RamRamez For review, I updated styles according to your feedback, please check it again

RamRamez (approved):

PR #805, opened 2022-05-25
Test new subgraph

aminlatifi:
Used test subgraph which product env is deployed to.


aminlatifi:
Let's check production with test subgraph and don't touch the main giveth subgraph. If everything goes well after the release, we can deploy new version of subgraph to Giveth main subgraph by merging https://github.com/Giveth/giv-token-subgraph/pull/61 and updating production.ts after they are fully synced.

MohammadPCh (approved):

PR #824, opened 2022-05-29
Update subgraph slice and add price

MohammadPCh:
- #720


src/features/price/price.services.ts
```diff
@@ -0,0 +1,49 @@
...
+		tokenId: tokenId.toLowerCase(),
+		daiId: '0x6b175474e89094c44da98b954eedeac495271d0f'.toLowerCase(),
+	};
+	const { data } = await gqlRequest(
+		config.MAINNET_CONFIG.uniswapV2Subgraph,
+		false,
+		query,
+		variables,
+	);
+	const givEth = new BigNumber(data.givtoken.derivedETH);
```
aminlatifi, line 30: Based on the query in `src/features/price/price.queries.ts`, it should be
```suggestion
	const tokenEthPrice = new BigNumber(data.token.derivedETH);
```

Also, as the aimed token is not always `GIV`, we can rename the variable name.

src/types/config.ts
```diff
@@ -98,6 +98,7 @@ export interface RegenStreamConfig {
 
 export interface BasicNetworkConfig {
 	TOKEN_ADDRESS: string;
+	tokenAddressOnUniswapV2: string; // For price purpose
```
aminlatifi, line 101:
```suggestion
	tokenAddressOnUniswapV2: string; // For price purpose in test env, on production this must have the same value of `TOKEN_ADDRESS`
```

aminlatifi (changes requested):
Thanks @MohammadPCh

aminlatifi (approved):

Issue #723, opened 2022-05-16
Remove Apollo from our application


MohammadPCh:
We use Apollo client only to communicate to the Graphql server and we don't use any feature of it. so I think It's better to remove it from our app and use fetch (The browser's built-in API) to communicate with the server.

MohammadPCh:
@mateodaza @RamRamez @alireza-sharifpour 
Dear Friends,
I am eager to know your comments.

RamRamez:
Yeah, we are not using its features. I checked all our graphQL end-points and I think we can't use Apollo cache feature in most of them because we need to update like count on user reaction. In some cases we need to send token authorization in header and cache is not useful either.
And lots of end-points are mutations that can be handled easily.

MohammadPCh:
Today we had a meeting with @alireza-sharifpour and have an agreement about the structure of our replacement. @alireza-sharifpour could you please add snippet code here?

alireza-sharifpour:
Yes, I agree. I think we can go with the browser’s native fetch and write a function or middleware for doing the repetitive stuff.
I will work on the function and send you the code to hear your opinions.

alireza-sharifpour:
Ok, Here is our helper function:
```ts
export function gqlFetch(
	query: string,
	variables: {} = {},
	additionalHeaders: HeadersInit = {},
) {
	const token = !isSSRMode ? localStorage.getItem(StorageLabel.TOKEN) : null;
	const Authorization = token ? `Bearer ${token}` : undefined;
	const defaultHeaders = {
		'Content-Type': 'application/json',
		...additionalHeaders,
	};
	const headers = Authorization
		? {
				...defaultHeaders,
				Authorization,
		  }
		: { ...defaultHeaders };
	try {
		const res = fetch(links.BACKEND, {
			method: 'POST',
			headers,
			body: JSON.stringify({
				query,
				variables,
			}),
		});
		return res;
	} catch (error) {
		return Promise.reject(error);
	}
}
```
What is your opinions?
@MohammadPCh @RamRamez @mateodaza @aminlatifi

alireza-sharifpour:
Also a simple asyncThunk with and without the function:

Without:
```ts
export const fetchUserByAddress = createAsyncThunk(
	'user/fetchUser',
	async (address: string) => {
		try {
			const userData = await fetch(links.BACKEND, {
				method: 'POST',
				headers: {
					'Content-Type': 'application/json',
				},
				body: JSON.stringify({
[...]

PR #833, opened 2022-05-30
Remove apollo client

MohammadPCh:
- #723


mateodaza (changes requested):
Awesome work man, but I found something.

The problem is with the token authentication and logout. I login with metamask but I seem to get the option to sign out even when I haven't signed anything which means I don't currently have a token.
Can you check it out? Everything else seems to work properly

image

mateodaza:
@MohammadPCh I solved the issue, it was a problem on my browser. SO, it's fine :)

Fix the conflict and we are ready to move on. Awesome work

MohammadPCh:
Thank you @mateodaza 
I think we can postpone it to after the verification and the GIVpower release.
Because we use Apollo on those branches and conflicts will be uncountable :D

Issue #745, opened 2022-05-19
Verification/Front-end: Main framework


RamRamez:
The general framework includes background, menus and main window

divine-comedian:
Is there anyway to test this issue as a user?

PR #864, opened 2022-06-05
Added Verification context and getting project data in it and created…

alireza-sharifpour:
#745


MohammadPCh (approved):
Thank you @alireza-sharifpour

Issue #880, opened 2022-06-06
Add donation failed messages


RamRamez:
I added different donation failed modal for different cases:
1. donations failed
2. donation is mined but not saved on our DB
3. User rejected the transaction!
4. donation cancelled
5. unknown error (is same as failed error)

divine-comedian:
testing on `staging.giveth.io`

on the `donation is mined but not saved on our DB` modal we tell the user to contact support, maybe it would be helpful to provide a direct link? 
![image](https://user-images.githubusercontent.com/67759413/172681170-e776f63a-e954-4fbc-b59f-76d6196afca9.png)

what do you think for sending them to the #support channel on discord?
https://discord.gg/TeWHtAjNQT
we can make this text `Please contact our support team.` into a link to the url above

error message `donations failed` looks good

same with `User rejected the transaction!` - I would only suggest updating the text to `Transaction was rejected!` since you are talking to the user. keep it consistent with other messages that "talk" to the user.

How can I test the donation error message for `donation cancelled`?

RamRamez:
@divine-comedian To test cancelled donation, you can try to donate and while your donation is pending, try to cancel it in MetaMask.

In PR #904:
1. I added discord support channel link to donation not saved message
2. changed rejected message to `Transaction was rejected!`

divine-comedian:
testing on `staging.giveth.io` 

![image](https://user-images.githubusercontent.com/67759413/173134247-69be1d2d-170d-49e8-9052-f3fcf4a089a1.png)

new link added looks good 

new rejected message looks good.

tx cancelled also appears to be working

![image](https://user-images.githubusercontent.com/67759413/173135547-58b4ffc8-0261-499f-a9a5-04b93ff44983.png)

moving to DONE

PR #881, opened 2022-06-06
Add donation failed messages

RamRamez:
Ref. #880 

I added different donation failed modal for different cases:

1. donations failed
2. donation is mined but not saved on our DB
3. User rejected the transaction!
4. donation cancelled
5. unknown error (is same as failed error)


alireza-sharifpour (approved):
Thanks @RamRamez

PR #888, opened 2022-06-07
update head and metadata

divine-comedian:
update meta and head data to show updates


MohammadPCh:
@divine-comedian I think we can close this PR, am I right?

divine-comedian:
@MohammadPCh did you end up merging the other one you made with ahmad's updates?

MohammadPCh:
@divine-comedian 
Yes, I did them all.
but we have a new route, `/givpower`.
we can add it later.

PR #928, opened 2022-06-15
add user token validation before donating (develop branch)

RamRamez:


mateodaza (approved):
we should think for later adding a custom modal for letting the user know why it was logged out or asking him to sign in again

PR #931, opened 2022-06-16
Add/save managing funds

alireza-sharifpour:


MohammadPCh (changes requested):
Thank you @alireza-sharifpour 
I think we need to have another select component for choosing a chain.

MohammadPCh (approved):
Thank you guys ;)

PR #941, opened 2022-06-16
Fixed add network on metamask

aminlatifi:
Updated gnosis-chain default nodeUrl parameter in config files


aminlatifi:
NOTE: current network switch cannot add network to metamask if it has not been added yet!

PR #943, opened 2022-06-17
Fixed add network on metamask

aminlatifi:
Updated gnosis-chain default nodeUrl parameter in config files


MohammadPCh (approved):
Thank you @aminlatifi

Issue #950, opened 2022-06-19
Add react hook form


MohammadPCh:
Use the `react-hook-form` for our forms. It makes everything about forms easier for us.

PR #951, opened 2022-06-19
Use react hook form

MohammadPCh:
- #950 
@mateodaza @RamRamez @alireza-sharifpour I'm eager to know your comments about this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
 		"react-dom": "^18.0.0",
 		"react-dropzone": "^11.5.3",
 		"react-google-maps": "^9.4.5",
+		"react-hook-form": "^7.32.2",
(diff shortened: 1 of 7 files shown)


RamRamez:
Very nice, thanks @MohammadPCh 
I think we can add real-time validation to improve UX.
Also I suggest to merge this branch to verification develop branch to prevent conflicts result from some changes like regex

MohammadPCh:
Thank you @RamRamez 
We can set our revalidation policy in each form. so it's ok I think.
Yeah I implemented it here first then I will merge develop to verification-develop-branch as always ;)

RamRamez (approved):
Thanks @MohammadPCh

alireza-sharifpour (approved):
Thanks @MohammadPCh

PR #968, opened 2022-06-21
update harvest modal

divine-comedian:


src/components/modals/HarvestAll.tsx
```diff
@@ -407,19 +407,28 @@ export const HarvestAllModal: FC<IHarvestAllModalProps> = ({
...
+													false,
+												),
+											) -
+												Number(
+													formatWeiHelper(
+														givBackStream,
+														config.TOKEN_PRECISION,
+														false,
+													),
+												)}
```
MohammadPCh, line 423: Thank you @divine-comedian 
I suggest this solution:
```
{formatWeiHelper(
  rewardStream.minus(
	  givBackStream,
  ),
  config.TOKEN_PRECISION,
  false,
)}
```

MohammadPCh (changes requested):

src/components/modals/HarvestAll.tsx
divine-comedian: thank you, made the changes!

Issue #981, opened 2022-06-22
Use React-Hook-Form in all Verification steps


alireza-sharifpour:
We should use React-Hook-Form in All Verification steps inputs

alireza-sharifpour:
Verification Steps:

- [ ] Personal Info
- [ ] Personal Social Media
- [ ] Project Registry
- [ ] Project Contact
- [ ] Activity And Milestones
- [ ] Managing Funds

PR #994, opened 2022-06-23
Feat/981 add react hook form to project registry

alireza-sharifpour:
#981


RamRamez (approved):
Thanks @alireza-sharifpour

Issue #997, opened 2022-06-23
Add Error Styles to text area component


MohammadPCh:
We need to have a component that handles errors and shows them like Input(if it has an error, change the border color and ...)

PR #1001, opened 2022-06-23
Verification/ add error styles to text area component

RamRamez:
Ref. #997 

add description validation to manage funds
add description validation to milestones
add validation to project registry


alireza-sharifpour (approved):
Thanks @RamRamez

PR #1013, opened 2022-06-27
Puts xDAI first on list after GIV

mateodaza:


MohammadPCh:
Thank you @mateodaza,
as we talked about in the weekly call, we gonna add an order value to the database, and sort tokens by that value.
so let the backend implement it then we can sort the array in the front-end with that.
Thanks.

MohammadPCh:
It fixed on #1018

PR #1015, opened 2022-06-28
added new styles for Select component and added registery name

alireza-sharifpour:
In updated design we need to add company name and image upload for verification's registry section
Also matched Select styling with the library


MohammadPCh (approved):
LGTM, thank you @alireza-sharifpour

PR #1028, opened 2022-06-29
add/handle status in done step

alireza-sharifpour:


src/components/views/verification/Done.tsx
```diff
@@ -4,30 +4,75 @@ import React from 'react';
...
+		],
+		[
+			EVerificationStatus.SUBMITTED,
+			`We received your application!
+		The Verification Team will send an email once it has been reviewed.`,
+		],
+		[EVerificationStatus.REJECTED, 'Please contact support team.'],
+		[EVerificationStatus.VERIFIED, ''],
+	]);
+
```
MohammadPCh, line 38: I think the object is good and we don't need Map here.

src/components/views/verification/TermsAndConditions.tsx
```diff
@@ -33,14 +33,30 @@ export default function TermsAndConditions() {
...
 			sendReq();
 		} else {
+			setVerificationData(prevState =>
+				prevState
+					? {
+							...prevState,
+							status: EVerificationStatus.DRAFT,
+					  }
+					: undefined,
+			);
```
MohammadPCh, line 59: I suggest you have a function and call it to avoid repeating the code. (I know it's not a lot of repetition, but we can reduce the number of lines and this is a suggestion)
Thanks

MohammadPCh (changes requested):
Looks Good, just 2 suggestions for the cleaner code base.

MohammadPCh (approved):
Thank you @alireza-sharifpour

PR #1052, opened 2022-07-04
Remove/linkedin from verification social profile

alireza-sharifpour:


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #901, opened 2022-06-08
Front-end Test


MohammadPCh:
We always read about writing tests for applications. and all of us agree that it a necessary for the back-end. but shall we have it for the front-end? what are its advantages? we need to think about it. maybe it help us to have an application with fewer bugs.
God knows!
What's your idea @MoeNick?

MoeNick:
Yeah functional tests works but needs lots of bandwidth to get benefited from. Let's have QA specialist first, then we can complete it after

mateodaza:
Let's share in this thread some ideas on which framework should we use!

@alireza-sharifpour did I hear in some call that you had some implementations before?

+ @MohammadPCh @RamRamez

mateodaza:
In this [article from the nextjs docs](https://nextjs.org/docs/testing) they have some examples for cypress, playwright, jest and vitest

alireza-sharifpour:
> 

Yes I had some implementations and read lots of documents before.
The PR for our project's configuration is here:
https://github.com/Giveth/giveth-dapps-v2/pull/1074
I'm going to share some resources in the next comment

alireza-sharifpour:
I investigated on Big projects code bases and read some docs and watched some courses.
We can use Jest + React Testing Library for Unit testing and Mocking and ...
React Testing Library is a great tool and it provides lot's of React and DOM related utilities.
https://testing-library.com/docs/react-testing-library/intro/
https://chrome.google.com/webstore/detail/testing-playground/hejbmebodbijjdhflfknehhcgaklhano?hl=en
Best course i found for testing:
https://testingjavascript.com/

Kent C dodds has great articles about testing on his website.

Also for E2E tests we can use Cypress which I'm not used it before but i see that some repos like Uniswap are using it and it seems perfect.
https://www.cypress.io/
[Uniswap Repo](https://github.com/Uniswap/interface)
[Pancakeswap Repo
](https://github.com/pancakeswap/pancake-frontend)

MohammadPCh:
Wow, thanks @alireza-sharifpour and @mateodaza 
It's great. I suggest that each of you undertakes a part of this research and demo it for the team( dev team + POs + testers)
what do you think?

alireza-sharifpour:
> Wow, thanks @alireza-sharifpour and @mateodaza It's great. I suggest that each of you undertakes a part of this research and demo it for the team( dev team + POs + testers) what do you think?

I think it's a good idea 👍

[...]

PR #1074, opened 2022-07-06
added jest basic setup and configuration

alireza-sharifpour:
- #901


MohammadPCh (approved):
Thanks @alireza-sharifpour, LGTM

PR #1078, opened 2022-07-06
some fixes on edit project

RamRamez:
Ref. #1079 

1. add loader component
2. add a loader to edit project
3. first check if user is signed in then check if user registered
4. add project not available page to edit project if project not found
5. auto close welcome modal when user singed in, in edit project view


alireza-sharifpour (approved):
Thanks @RamRamez

PR #1083, opened 2022-07-07
Verification/bind checked steps to DB

RamRamez:
1. bind checked steps to last step in DB
2. prevent user to select any steps on mobile


alireza-sharifpour (approved):
Thanks @RamRamez

PR #1091, opened 2022-07-12
Added framer-motion to modals

alireza-sharifpour:
@RamRamez @MohammadPCh


MohammadPCh:
@alireza-sharifpour  I think we can close this, am I right?

PR #1137, opened 2022-07-17
Fix/givpower explanatory responsive

alireza-sharifpour:
- #738


MohammadPCh (approved):
Thank you @alireza-sharifpour

Issue #1099, opened 2022-07-12
Verification Form - Required Questions


WhyldWanderer:
This question should be required but currently it is not.

![image](https://user-images.githubusercontent.com/74267818/178496893-aa7d70b6-7acb-47c8-8d52-b60e74a9e0a5.png)

WhyldWanderer:
This one should also be required
![image](https://user-images.githubusercontent.com/74267818/178813919-6dbf94fe-c3a9-499f-9701-a3016e95ec87.png)

divine-comedian:
testing on `giv-verification.vercel.app/`

This is now a required field, I cannot proceed until I choose one.
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/67759413/179579590-a3c72065-9fd5-47de-ae1a-addb736a8d4a.png">

and on the next `activity and milestones` page these fields are required, including the one specified in the issue:
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/67759413/179579897-75a012a8-5604-4373-8b47-e3a5d6b507a5.png">


issue seems to be resolved! moving to DONE

PR #1141, opened 2022-07-17
Handled Required Fields

alireza-sharifpour:
- #1099


MohammadPCh (approved):

Issue #1095, opened 2022-07-12
Verification Form - Personal Social Media content


WhyldWanderer:
The content on the personal social media section does not reflect the figma. Also, this should be optional.
![image](https://user-images.githubusercontent.com/74267818/178491493-8f048860-8c54-4512-9290-e2354fad83fa.png)

It should say "Connecting your social media accounts is a good way to show that
you are who you say you are." As reflected below.
![image](https://user-images.githubusercontent.com/74267818/178491654-2927fa8a-26f2-46ff-976f-f95fb4c7060e.png)

Thanks!!

PR #1143, opened 2022-07-18
Changed Social profile description

alireza-sharifpour:
- #1095


MohammadPCh (approved):
Thanks @alireza-sharifpour

PR #1154, opened 2022-07-19
open stake modal by query param

MohammadPCh:
- #1145 

try it out like this:
/givfarm?open=`[StakeingType]`&chain=`[chainId]`

staking types:
- UniswapV2_GIV_DAI
- UniswapV3
- Balancer
- Sushiswap
- Honeyswap_GIV_HNY
- Honeyswap_GIV_DAI
- GIV_LM
- Honeyswap_FOX_HNY
- UniswapV2_CULT_ETH

chains:
- gnosis
- main

so if you want to show the GIV stake modal on Gnosis, the address will be something like this:
/givfarm?open=GIV_LM&chain=gnosis

PR #1158, opened 2022-07-19
Optimized open staking card directly

aminlatifi:
@MohammadPCh #1154 seems fine, just made this PR to improve user experience

### DIFF
--- a/src/components/cards/BaseStakingCard.tsx
+++ b/src/components/cards/BaseStakingCard.tsx
@@ -73,6 +73,7 @@ import { useAppDispatch, useAppSelector } from '@/features/hooks';
 import { ITokenDistroInfo } from '@/types/subgraph';
 import { switchNetwork } from '@/lib/wallet';
 import { setShowWalletModal } from '@/features/modal/modal.slice';
+import Routes from '@/lib/constants/Routes';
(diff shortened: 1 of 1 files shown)


src/components/cards/BaseStakingCard.tsx
```diff
@@ -179,20 +180,19 @@ const BaseStakingCard: FC<IBaseStakingCardProps> = ({
 				: config.MAINNET_NETWORK_NUMBER;
 		const checkNetworkAndShowStakeModal = async () => {
 			if (_chain && _open === type) {
-				await switchNetwork(_chainId);
+				dispatch(setShowWalletModal(!isWalletActive));
+				if (isWalletActive) await switchNetwork(_chainId);
```
MohammadPCh, line 184: It calls the switchNetwork function immediately after showing the wallet modal. so I think It will be buggy.

MohammadPCh (changes requested):

src/components/cards/BaseStakingCard.tsx
aminlatifi: will `isWalletActive` be positive?

MohammadPCh: Yeah... you are right ;)

Issue #1057, opened 2022-07-04
GIVpower - APR should be dynamic


laurenluz:
This wasn't yet clear in the figma, but the GIVpower APR will be different for each person.
- [x] if the user has nothing staked/locked, the APR should show a range from min to max like (120%-5090%)
- [x] if the use has something staked/locked, the APR shown should be their average APR, considering all staked/locked positions

@mosaeedi is working on update this in the figma and can add the link here when it's done

@MohammadPCh can you take this one on?

![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/c485dd81-a5f0-4a76-8d54-c4810884dc20)

laurenluz:
Hey @MohammadPCh There is an issue w/ this one:

You can see that I an connected w/ my wallet and staking... so I should be seeing *my* average APR, not the full range, as it is shown here:

![image](https://user-images.githubusercontent.com/75490971/179509028-ca0ff62a-7d28-4ad7-a2e6-f2d1eaaf73ac.png)

The user should be seeing the "range" only if they are not staking GIV, or if their wallet is not connected.

laurenluz:
@MoeNick this one has a bug, needs to be fixed again

MohammadPCh:
Hey @laurenluz 
I show the average when the user has some locked GIV, and if the user hasn't any locked give I show him the apr range.
Is it wrong?

MohammadPCh:
Hey @laurenluz.
the GIVpower is on the gnosis chain, and if the user's wallet is not connected we shouldn't show the GIVpower Card (I know we show it now, and it's a bug, we should show the old GIV staking card or decide about the bridge).
If we change our default chain to the Gnosis chain, we have another problem. consider we show the GIVpower APR range but other pools are zero. because we need the network provider for calculating their APR. I suggest it's not a good idea to show one pool APR's correct and zero for the rest it may confuse users.
What do you think?

Cotabe:
@laurenluz @MoeNick @MohammadPCh 

If nothing is staked it shows the range.
If there is something stacked it shows the value.

I'm not sure I understand well the latest problem

laurenluz:
@Cotabe That's awesome.

To test the latest problem... basically, you need to see what it looks like if the user's wallet is not conencted. If they are not connected, we wanted to set it up so that the default view is gnosis chain... and that the user *can* see the APRs...

previously, under these conditions, the user could not see APRs.

PR #1163, opened 2022-07-19
Fix apr range display condition

MohammadPCh:
- #1057


aminlatifi:
@MohammadPCh It shows zero for me when the wallet is not connected

image

MohammadPCh:
@aminlatifi 
Thanks for your notice, But I will fix it in another PR because It needs more communication with @laurenluz.
the GIVpower is on the gnosis chain, and if the user's wallet is not connected we shouldn't show the GIVpower Card (I know we show it now, and it's a bug, we should show the old GIV staking card or decide about the bridge).
If we change our default chain to the Gnosis chain, we have another problem. consider we show the GIVpower APR range but other pools are zero. because we need the network provider for calculating their APR. I suggest it's not a good idea to show one pool APR's correct and zero for the rest it may confuse users.
I will post this comment to the issue to continue there.

aminlatifi (approved):

PR #1166, opened 2022-07-20
$nice landing banner

mateodaza:


MohammadPCh (changes requested):
Thanks @mateodaza 
Based on the design I think we need to use the Container component for it.
Screenshot from 2022-07-20 10-24-21

Issue #1068, opened 2022-07-05
All wrong project slugs are redirected to maintenance page


RamRamez:

jeevanplus007:
@RamRamez Are there different types of Wrong Projects ? Which scenarios are covered as part of Fixing the issue ?

divine-comedian:
testing on `staging.giveth.io`

test cases:
while logged in with metamask
#1 - shows this:
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/67759413/179846762-d18f16bd-59d3-483e-99be-ebbdb53471b8.png">

#2 shows 404 page
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/67759413/179846848-c5a4b180-c4f9-4263-a1e1-1bfc10fe9aa2.png">

#3 and #4 show 404 page as well.

#5 - while signed in on an account other than the project owner I see this 
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/67759413/179847263-1b5223bb-3e14-4f36-b2ab-23806282fa30.png">

#6 while connected with the project owner wallet I see my draft preview
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/67759413/179847994-0c1197d8-848d-44a1-9ad8-6075eb3fd732.png">

#7 cancelled projects cannot be found, even by the user that created it 
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/67759413/179849408-61983b27-a30c-4382-b374-d1ad9ada9ff5.png">

--
# There is only one issue...

if I am not signed into metamask, any of the above scenarios (with the exception of the 404 pages) only show me a blank project screen:
![image](https://user-images.githubusercontent.com/67759413/179849840-5ea66c1d-0b58-45e5-bbac-8ccf25b94144.png)

If I visit a project that is unaccessible or I have the wrong-slug it should show me the wrong slug screen instead of a blank project screen.

divine-comedian:
moving this to sprint backlog, maybe take a look @MoeNick

RamRamez:
thanks @divine-comedian for complete testing and reporting.
Solved in PR #1178

divine-comedian:
testing again on `staging.giveth.io`

I test any of the scenarios above with my metamask locked(not signed in). I am now directed to the project not found page instead of a blank project!
[...]

PR #1178, opened 2022-07-20
fix showing blank page when wallet is not connected

RamRamez:
#1068


MohammadPCh (approved):
Thanks @RamRamez

MohammadPCh (approved):
thanks @RamRamez

Issue #1106, opened 2022-07-13
Create a Project - Upload Image Cover - Not properly refreshing the removed data


jeevanplus007:
**Issue :** Create a Project - Upload Image Cover - Not properly refreshing the removed data. 

**Steps to Repro :**
1. Click on Create a Project
2. Go to Add an image to your project
3. Click on Upload Cover Image
4. Using Drive, Upload any image from System(Hard drive)
5.  After Successfully Uploading, Click REMOVE button
6. Now Select any image from our gallery as shown picture

**Results :** The Image Details of previously Uploaded information is displayed which is an issue. 

**Snapshot 1 for Reference :** Upload Image Issue 
![image.png](https://images.zenhubusercontent.com/62a3cf4962612d632dd3b42a/7cbb2e19-ba46-48c6-9ed8-783838270dbd)

oyealmond:
Tested in https://staging.giveth.io/create 
After following the steps (upload a picture from Hard drive, removing it, and then upload a picture from Giveth's gallery), the past image information is not shown:

<img width="737" alt="Captura de Pantalla 2022-08-12 a la(s) 11 06 38" src="https://user-images.githubusercontent.com/110692280/184396971-9eefa53f-3b41-494c-999e-df6e52bcba7b.png">

PR #1179, opened 2022-07-20
fix Not properly refreshing the removed image in create project

RamRamez:
#1106


MohammadPCh (changes requested):
Thanks, @RamRamez ,
It's not a good idea to have 2 states in the parent component for one purpose.
If you only want to fix this issue, you can remove the ImageUploader component when the user selects form background images.
also, we have the FileUploader component, which I pass URL to file there. I think maybe it's better to change it to a general file uploader (I didn't have enough time at that time) but if it's now urgent I can change it.

MohammadPCh (approved):
Thanks @RamRamez ,
We need to merge this and file uploader component together.

PR #1194, opened 2022-07-26
Updates what is stream modal and text missing on signin modal

mateodaza:


src/components/modals/WelcomeModal.tsx
```diff
@@ -113,6 +113,7 @@ const WelcomeModal: FC<IModal> = ({ setShowModal }) => {
 };
 
 const ModalGrid = styled.div`
+	color: black;
```
MohammadPCh, line 116: @mateodaza please use our design system colors, I think it's neutralColor.gray[100]

MohammadPCh (changes requested):
Thanks, @mateodaza, just a small fix is required.

MohammadPCh (approved):
Thanks @mateodaza

PR #1212, opened 2022-07-27
verification / force to send data on terms step

RamRamez:
1. force to send data on terms step
2. when submitting form on last step, show error instead of showing done page


MohammadPCh (approved):
LGTM, Thanks @RamRamez

PR #1217, opened 2022-07-27
verification / fix progress-bar width

RamRamez:
progress-bar width should be full when form is submitted

image


alireza-sharifpour (approved):
Thanks @RamRamez

Issue #1025, opened 2022-06-28
GIVbacks round number and dates are wrong


laurenluz:
When my walllet is not connected, I'm seeing this:

![image](https://user-images.githubusercontent.com/75490971/176226028-d0658133-493b-4906-88cb-e784a45b0b12.png)

It should say GIVbacks Round 14... Also the dates/times are wrong. When I connect my wallet, it shows up correctly:

![image](https://user-images.githubusercontent.com/75490971/176226265-1fec59af-9c45-42a0-8fa0-edd59c348837.png)

divine-comedian:
testing on `staging.giveth.io` 
still though something seems wrong 
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/67759413/182456138-b36083f4-249c-41d4-8549-439d2a2c4a9f.png">




the start date and end date show the round begins and ends on thursday @ 9.10am - it should be friday @ 10am according to my local time

MohammadPCh:
@divine-comedian thanks for testing. do you test it in the production environment?

divine-comedian:
Looks to be working on `giveth.io`

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/67759413/182648096-33e6ff42-25ff-4e65-86cf-9eb7db0eeb20.png">

It would be helpful next time to know where to test when this issue is moved to UAT

Hoping to see those responsiveness changes merged soon. :)

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/67759413/182648337-c0945e73-ca99-40ac-a1ce-72a6022f352a.png">

moving to DONE

PR #1223, opened 2022-07-27
get subgraph data even no connected wallet

MohammadPCh:
- #1025


src/lib/subgraph/subgraphQueryBuilder.ts
```diff
@@ -165,9 +176,10 @@ export class SubgraphQueryBuilder {
 	};
 
 	private static generateFarmingQueries = (
-		userAddress: string,
 		configs: Array<SimplePoolStakingConfig>,
+		userAddress?: string,
 	): string => {
+		if (!userAddress) return '';
```
aminlatifi, line 182: We need unipool info for calculating APR

aminlatifi: even if the user's wallet is not connected

aminlatifi (changes requested):

Issue #1201, opened 2022-07-26
I can no longer see the wallet address of the project


WhyldWanderer:
I can no longer see the projects wallet address on the Donations tab of their project's page.

![image](https://user-images.githubusercontent.com/74267818/181095027-37bb3e67-f1f9-460e-8156-18abd6d6c2a7.png)

Now that we have the functionality to have 2 wallet addresses, we should show both of them here.

MoeNick:
Sorry we will fix this asap

mosaeedi:
Here is how it should look like,

Single address on both networks: [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=11775%3A207600)

Different addresses for both networks: [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=11775%3A208513)

MoeNick:
@mateodaza Accepted to take this one

mateodaza:
Merged as hotfix to live version

WhyldWanderer:
that was fast!! It looks good Thanks Mateo!

RamRamez:
Thanks @mateodaza. Please link the related PRs to issues. It helps us track changes.

divine-comedian:
@WhyldWanderer is this issue tested successfully? can it be moved to done?

MoeNick:
Moved to production as hotfix.

MoeNick:
Moved to production as hotfix.

PR #1234, opened 2022-07-28
Remove project walletAddress

RamRamez:
Ref. #1201


mateodaza (approved):
Thanks a lot for your refactor on multiple addresses, this is looking and working great

I just found a tiny bug that I think we can add in here 

image


I made a donation with DRGIV which doesn't have a value and the ProjectTotalFund card won't show, this is because we are looking at the total USD donated, I believe we should look on the number of donations made instead. 

I'll add this fix on another commit :)

Issue #1124, opened 2022-07-14
SingleProject Page: Refactoring showing categories


MoeNick:
Project Single Page - how to show categories / subcategories? with links? 

## Design resource

Project single page **[Owner view]** -> [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=11651%3A208431)

Project single page **[Public view]** -> [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=11651%3A208400)

![Screen Shot 2022-07-27 at 4.09.11 PM.png](https://images.zenhubusercontent.com/61dd85b9a1419a0fc83f15d1/b4409495-eb19-49c0-ab8a-475e760a6d93)

maryjaf:
On what environment can I test this task?

MoeNick:
https://givfilter.vercel.app
@maryjaf 
This is the feature branch for all issues in the Epic

maryjaf:
In addition to the problem of displaying sub-categories horizontally, I encountered this problem during the test
I have a project whose main category is environment&energy and subcategory is air
When I search in this category and subcategory, it shows the name of the project(testimer)
But I edit the project and add a  new sub-category(energy)
Neither the previous sub-category(air) nor the new sub-category(energy) will be displayed in the search
In fact, it seems that there is a problem with the project and filter editing scenario

MoeNick:
I can see your project under both sub-categories. 
![image](https://user-images.githubusercontent.com/80677512/188307301-1b2072c6-7c95-4662-bb00-fd5b6cabbc6b.png)

![image](https://user-images.githubusercontent.com/80677512/188307351-a56a9c63-d989-4219-8b28-a4e2b6d404a3.png)

Can you reproduce it again @maryjaf ?

maryjaf:
I got it 
When I edit the project, the status of the list changes to "Waiting for review"
And that is why it is not displayed in the search engine @MoeNick

PR #1246, opened 2022-07-31
1124/show main categories

alireza-sharifpour:
- #1124


src/helpers/projects.ts
```diff
@@ -0,0 +1,16 @@
+import { ICategory, IConvertedCategories } from '@/apollo/types/types';
```
RamRamez, line 1: since this file is not related to projects page, I think you can rename it to 'singleProject.ts'

RamRamez (changes requested):
Great thanks @alireza-sharifpour. Just needs a tiny fix.

alireza-sharifpour:
Thanks @RamRamez for your attention, I fixed the things you mentioned.

RamRamez (approved):
Thanks @alireza-sharifpour

PR #1251, opened 2022-08-01
Fetch unipool info from chain

aminlatifi:
Fetch unipool info from the chain in case subgraph fetch was not successful


MohammadPCh (approved):
Thanks @aminlatifi

PR #1252, opened 2022-08-01
Refactor unipool info fetching in APR calculation

aminlatifi:
Moved fetching unipool info to a function for all APR calculation


MohammadPCh (approved):
LGTM, thanks @aminlatifi

Issue #1121, opened 2022-07-14
Filtering/Sorting


MoeNick:
@mosaeedi commented on [Tue Jul 12 2022](https://github.com/Giveth/GIVeconomy/issues/658)

![Screen Shot 2022-07-12 at 3.50.16 PM.png](https://images.zenhubusercontent.com/61dd85b9a1419a0fc83f15d1/fa49cb5e-99f2-4547-a87b-f9d64920e4d1)

## Design resource
Filter/sort module -> [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=11069%3A199976)

**Note:**
The filter and sorting module should work both on the categories and search results.
---

divine-comedian:
Looking at the current filter views there's a few things that don't quite make sense to me.

As a user from the four filter options presented if I select GivingBlock and Accepts GIV token will always yield no results. We could assume any project that isn't a giving block project would accept GIV tokens at the moment, right? 

The Giving Block also already has its own sub category under non-profit so having this as a filter might seem redundant. 
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/67759413/184698538-446c7d14-b4b6-443c-82f8-82261dceafcd.png">

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/67759413/184698595-26ac77a3-a250-4728-ba54-3421d12254c3.png">


we should also consider removing "from Trace" since Giveth Trace is no longer maintained. 

The user should have the ability to 'clear filters'

maybe some suggestions for different filters:

`Has updates` (the project has at least 1 update)
OR 
`Recently Updated` (where we define recent as < 3 months)

perhaps also a 'least funded' option for the sort.

MoeNick:
- [ ] we should also consider removing "from Trace" since Giveth Trace is no longer maintained. 

- [ ] The user should have the ability to 'clear filters' - I'm also thinking like that, @mosaeedi can you take a look? 

@divine-comedian for your suggestion, due to it might have spec changes, I'll put them in another issue, let's focus on functionality in this step.

MoeNick:
Unfortunately givingblock has different labels on staging and I can say also in production, so we can't rely on that fact.  @divine-comedian

divine-comedian:
> Unfortunately givingblock has different labels on staging and I can say also in production, so we can't rely on that fact. @divine-comedian

[...]

PR #1257, opened 2022-08-01
1121/add filter logic

alireza-sharifpour:
- #1121


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #1188, opened 2022-07-25
GIVpower - add "lock" icon & tooltip to the GIV amount under "unstake"


laurenluz:
*If* the user has some part of their GIV locked, we shoud show a "lock" icon (where i circled below).

This lock should be a tooltip that says "Some or all of your staked GIV is locked. Click "Locked GIV details" for more information."

![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/33aa3272-78bf-4e5a-82e6-b66f12c7f1d4)

@MohammadPCh or @mateodaza - can you take this?

laurenluz:
The issue this is solving is the fact that the number under "unstake" is the total amount of staked GIV... although some may be locked and not able to be unstaked. I noticed this and the @Cotabe made a separate issue about it as well (independent of my influence).

I do not think we should link the user directly to the modal... this is even more confusing. they're just trying to figure out what the unstaked tokens are at this time.

There will be a link to that modal at the bottom of the card.

Regarding the "lock" symbol of the "?"... I mean, if @mosaeedi doesn't like the lock and insists on the "?", then ok... but I think the "lock" showing up only when some part of the tokens are locked actually provides information at first glance... whereas the "?" only seems to add more confusion... like why should you by wondering about the number under unstake?

divine-comedian:
still waiting for the lock icon from @mosaeedi to move forward on this.

mosaeedi:
Since we are going to show a tooltip, I think it's better to show the "?" icon because we always used the "?" for the tooltip and when users saw it, they know there is some more information, but showing the lock, may actually prevent user hover on the icon and instead they may think all of the tokens are locked!

laurenluz:
ok! thank you @mosaeedi 

@MohammadPCh - can you implement?

divine-comedian:
pending in PR 
https://github.com/Giveth/giveth-dapps-v2/pull/1274

divine-comedian:
Ready to test on:
https://givpower.vercel.app/givfarm

WhyldWanderer:
Testing on https://givpower.vercel.app/

I see the ? and the following message:
![image](https://user-images.githubusercontent.com/74267818/182913766-6a6cbff9-4333-444e-ada4-6de36fbc1194.png)


Looks good! Moving to done

PR #1274, opened 2022-08-02
Locked icon tooltip

divine-comedian:
features requested in https://github.com/Giveth/giveth-dapps-v2/issues/1188

### DIFF
--- a/src/components/cards/BaseStakingCard.sc.tsx
+++ b/src/components/cards/BaseStakingCard.sc.tsx
@@ -13,7 +13,6 @@ import {
 } from '@giveth/ui-design-system';
 import { Flex } from '../styled-components/Flex';
 import { Shadow } from '@/components/styled-components/Shadow';
-import { IconWithTooltip } from '@/components/IconWithToolTip';
 
 interface IStakingPoolContainer {
 	shadowColor?: string;
@@ -205,11 +204,6 @@ export const DisableModalCloseButton = styled(OulineButton)`
 export const DisableModalImage = styled.div`
 	width: 36px;
 	color: ${brandColors.giv[500]};
-`;
-
-export const LockIconTooltip = styled(IconWithTooltip)`
-	padding: 10px !important;
-	margin-left: 10px !important;
 `;
 
 export const LockInfotooltip = styled(Subline)`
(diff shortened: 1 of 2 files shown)


src/components/cards/BaseStakingCard.sc.tsx
```diff
@@ -88,8 +88,10 @@ export const StakeContainer = styled(Flex)`
 	gap: 6px;
 	width: 100%;
 `;
-export const StakeAmount = styled(Caption)`
-	text-align: center;
+export const StakeAmount = styled(Flex)`
+	align-items: center;
+	justify-content: center;
+	font-size: 0.9em;
```
MohammadPCh, line 94: Thanks @divine-comedian 
FYI we have a styled component called FlexCenter that can use instead of this.
also for `font-size: 0.9em` we don't have this in our design system, I prefer this approach:
```tsx
<FlexCenter>
    <StakeAmount></StakeAmount>
    {isLocked && <IconWithTooltip />}
<FlexCenter>
```

src/components/cards/BaseStakingCard.tsx
```diff
@@ -517,12 +517,27 @@ const BaseStakingCard: FC<IBaseStakingCardProps> = ({
...
+									<StakeAmount gap='5px'>
 										{type === StakingType.UNISWAPV3_ETH_GIV
 											? `${stakedLpAmount.toNumber()} ${unit}`
 											: `${formatWeiHelper(
 													stakedLpAmount,
 											  )} ${unit}`}
+										{isLocked && (
+											<IconWithTooltip
+												icon={<IconHelp size={16} />}
+												direction={'right'}
```
MohammadPCh, line 529: add right direction in that position, may cause adding horizontal scroll to the page like this:
Screenshot from 2022-08-03 14-05-06
I suggest change it to left

MohammadPCh (changes requested):

src/components/cards/BaseStakingCard.tsx
divine-comedian: ahh! good to know, changed it to top so it doesn't block the text the tooltip is telling you to click :)

src/components/cards/BaseStakingCard.sc.tsx
divine-comedian: great! thank you for the suggestion

MohammadPCh (approved):
wow, thanks @divine-comedian

Issue #1228, opened 2022-07-27
Angel Vault Farm Card Frontend minor fixes


laurenluz:
Ok I talked it over with Griff, and to be more consistent with the other farm cards, suggesting a few small fixes to the frontend for the Angel Vault.

1. Angel Vault icon should be at the top next to "Angel Vault" text
2. We should put the two token logos... let's use our normal GIV logo, & the dark colour oneGIV logo that @aminlatifi created
3. Add a tooltip that says "Your cumulative APR is the sum of rewards earned as fees & added automatically to your position (x % IRR), and rewards earned in GIV from staking your LP (y% APR)." ... We will need to show the IRR% & the other staking % @aminlatifi 

![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/543aa28e-8c19-483c-94dd-4d98c5088065)

oneGIV icon:

![logo_256_onegiv (1).svg](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/e868939a-0843-4caa-9b4d-c1ded90b8266)

aminlatifi:
@laurenluz for the icon size and style we may need design. Can we have it?

MohammadPCh:
@laurenluz @mosaeedi 
please look at this:
![Screenshot from 2022-07-28 11-25-29](https://user-images.githubusercontent.com/17097576/181440759-ded22f51-7cf7-40d8-bdca-3b2117006fba.png)
is it acceptable?

Notice: The icon sizes are the same and 40px but because of colors and borders the oneGIV icon seems smaller.

laurenluz:
@mosaeedi i made a rough version of this in the figma on the angel vault frame. Can you help clean it up and like it here for the devs?

mosaeedi:
![Screen Shot 2022-07-28 at 10 14 58 AM](https://user-images.githubusercontent.com/13708322/181444207-58970e21-ab09-4d1a-a77d-78c960647c4c.png)


Here is the oneGIV/GIV card on Figma -> [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=11246%3A204389)

aminlatifi:
@mosaeedi @laurenluz I don't know where to put the APR tooltip!

aminlatifi:
@laurenluz @mosaeedi I don't know where to put the APR tooltip!
[...]

PR #1293, opened 2022-08-03
add tooltip and flex

divine-comedian:
updating apr tooltips as per
https://github.com/Giveth/giveth-dapps-v2/issues/1228


MohammadPCh (approved):
Thanks @divine-comedian 
I increased the gap on your behalf.

Issue #1280, opened 2022-08-02
GIVpower Page - when user's wallet is not connected, it should ask them to connect/switch network


laurenluz:
Right now, when the wallet is not connected, I see this:
![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/eba00bbf-e339-4b14-92d6-92171262d025)

Instead, it should prompt me to connect my wallet... Should look like this & say:
- [ ] Text: To see your GIVpower, please connect your wallet.
- [ ] Button: Connect Wallet
![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/a6500732-de09-463a-80a7-668afc173a4d)

Also, when my wallet is connected, the button should say:
- [ ] Boost Projects
(with an "s")

laurenluz:
nice! working great.

PR #1299, opened 2022-08-03
Fix givpower reward card

MohammadPCh:
- #1280


mateodaza (approved):
Thanks for making the styling better, tested and it's working properly. lgtm!

PR #1301, opened 2022-08-03
fix single project responsive issues

RamRamez:


mateodaza (approved):
Beautiful! nothing to add here

BEFORE
image

AFTER

image

Issue #1259, opened 2022-08-01
donation search only showing one page


divine-comedian:
testing `giveth.io`
brave browser clean cache and cookies

while attempting to look up all donations matching a certain token type I was not able to see donations made past the first "page"

normally while looking at a project's donations received the user has multiple pages they can go through to look at all donations from the most recent, going back.

however whenever I enter a search query the element to scroll through pages dissapears, this means I cannot really see all donations made to that project that match my search query.

Here's a quick loom video to explain further my issue:

https://www.loom.com/share/a9415a045195493fa2f0a2c1665ff68c

divine-comedian:
testing on https://staging.giveth.io/project/how-many-photos-is-too-many-photos
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/67759413/184407721-7e17d4c8-d16b-468e-97a9-25a371f62d7c.png">

I can now enter a search on the donations tab and can see multiple pages matching by search query. 

issue looks to be resolved - since I tested this on staging I will move this directly to DONE!

PR #1325, opened 2022-08-08
Fix/project donations search

alireza-sharifpour:
- #1259


RamRamez (approved):
Thanks @alireza-sharifpour

PR #1342, opened 2022-08-10
unify all device detect hooks

RamRamez:
Ref. #1351


MoeNick:
Would u pls connect to the issue?

RamRamez:
> Would u pls connect to the issue?

I created an issue for it (#1351).

MohammadPCh (approved):
Thanks @RamRamez , LGTM.

Issue #1343, opened 2022-08-11
Test issues on Givfilter


MoeNick:
testing on `https://givfilter.vercel.app/projects`
test on brave browser and firefox

I can test a combination of scenarios add a sorting and a combination of different filters. the project page updates and shows me the correct projects depending on the options I chose. 

I notice there are two strange bugs:

**The first**, sometimes on **firefox** when I log in while browsing projects I can hit the bottom of the list and there is still a load more button, clicking load more does not reveal any new projects and I know that I am at the bottom of the list according to my sorting/filter combination.

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/67759413/183751409-3fe3d531-c70e-469f-a500-e6297efb5c7d.png">

I can fix this by refreshing page after I sign in, it no longer shows this erroneous load more button while filtering/sorting

---

**The second** issue is that while signed in on either firefox and brave. I am already signed in and if I try to  apply two filters when I click load more the button shows the loading animation but does nothing. 

I can apply just one filter and the load more button works

If I am signed out of my web wallet I can click load more with two filters and it works.

If I sign out, refresh the page, sign in and apply two filters it works. If I refresh the page after this scenarion then I arrive at the same problem with the load more button.

I make a loom video to try and show off the bug better. 
https://www.loom.com/share/dca68f63cb3f4fdc82e8e590a0b427d7

_Originally posted by @divine-comedian in https://github.com/Giveth/impact-graph/issues/579#issuecomment-1209834528_

alireza-sharifpour:
These issues should be fixed now
https://givfilter.vercel.app/projects

@divine-comedian @MoeNick

franculio:
Tested on https://givfilter.vercel.app/projects with brave browser. It all seems to be working good!

![Captura de pantalla 2022-08-12 a las 17.22.06.png](https://images.zenhubusercontent.com/62ea6f3faed0f381f5fe2f36/d51eee9a-f7ed-47a7-a37b-e133c686063d)

![Captura de pantalla 2022-08-12 a las 17.22.37.png](https://images.zenhubusercontent.com/62ea6f3faed0f381f5fe2f36/93fbf81b-30f3-4d4b-b4a0-7328d7af38ad)

PR #1344, opened 2022-08-11
Fixed LoadMore Issue

alireza-sharifpour:
- #1343


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #1185, opened 2022-07-25
UX improvements of H-cases of multi-addresses. 


MoeNick:
Here is the design for all other cases, please take a look

[Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=11580%3A205321)

_Originally posted by @mosaeedi in https://github.com/Giveth/giveth-dapps-v2/issues/1150#issuecomment-1192624026_

divine-comedian:
testing on https://staging.giveth.io/

I am testing the H-cases (edge cases?) as stated in the figma above

when I enter an address into a field and then delete it or if I just leave it blank,  under the empty box it says "Eth adderess invalid" but the box is empty so the message should say "this field is required" right?
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/67759413/184401149-efc3af6b-9b91-465f-8149-92a7c1c27853.png">

according to the figma there should be a state where the user does not see the text  "You can enter a new address to receive funds on Mainnet network(or Gnosis Chain)." However I try to enter a new address or change the address while in create or edit project and it isn't clear what I should do as a user to make that message disappear. I don't think this is working.

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/67759413/184403723-f2c3d21a-321d-42cd-b7b6-ab247a558a13.png">

I think this issue still needs some fixes, moving back to sprint backlog

RamRamez:
For testers:
In PR #1372 
1. Fade-in and fade-out animation is added to the `resolved address` toast massage and to the `wallet address already used` error message
2. `This field is required` error message is added when wallet address input is empty

It is not clear for me when the caption `You can enter a new address to receive funds on Mainnet network(or Gnosis Chain).` should be shown. I asked mosaeedi and I'm waiting for his response.

PR #1372, opened 2022-08-15
add animation to multi-addresses

RamRamez:
Ref. #1185 

1. Fade-in and fade-out animation is added to the resolved address toast massage and to the wallet address already used error message
2. This field is required error message is added when wallet address input is empty


MohammadPCh (approved):
Thanks @RamRamez

Issue #1181, opened 2022-07-21
Some issues on verification - Manage funds - Add an address modal


RamRamez:
1. when title is typed, the counter is not working
2. If user is on a network other than Mainnet, app should notify user to change network, but it says `The address is not valid`
![image](https://user-images.githubusercontent.com/73441857/180278207-a4793514-5804-4eea-9b50-29170078e4cc.png)

3. If title is not unique, app should show error in input box not a toast error
![image](https://user-images.githubusercontent.com/73441857/180279160-c4ded27a-5262-4194-b061-d6c5e334a1c1.png)

4. App should prevent repetitive addresses
![image](https://user-images.githubusercontent.com/73441857/180279656-82f20119-7aba-44fa-9ed9-c7b5c694d798.png)

5. ENS addresses are not accepted when clicking next
![image](https://user-images.githubusercontent.com/73441857/180283661-e692a332-acc8-4b59-9dcb-9c2d64c3fcc0.png)

6. If network is not selected, app should show the error in the select component not a toast error
![image](https://user-images.githubusercontent.com/73441857/180299033-ff40a46c-ccbd-4d1a-b6e5-38e91a43fb3b.png)

divine-comedian:
testing on `staging.giveth.io`
brave browser
going through the issues by number:
1. counter appears to be working, I cannot type a title longer than the counter limit
2. using ENS on gnosis chain I get a small red message underneath asking me to change network, changing network resolves the red text
3. I cannot add a duplicate wallet title, I receive an error message, looks good
4. I cannot add a duplicate ENS name to the same chain
## however 
If I use a regular eth address I can add as many duplicates as I want no matter the chain selected! This looks like a bug -
<img width="1440" alt="image" src="https://user-images.githubusercontent.com/67759413/184887024-8277cb09-69fc-4d44-901f-c4f79eb43919.png">


5. ENS addresses render as regular eth addresses and are accepted when I click next.
6. When I don't select a chain I am thrown an error and not allowed to proceed. this looks good.

divine-comedian:
testing this again on `staging.giveth.io` 

now if I try to add the same eth address on either gnosis chain or ropsten I receive an error text underneath the input box and I am prevented from adding this address. 
![image](https://user-images.githubusercontent.com/67759413/184930974-42fed47f-c0d2-4602-9950-52f374160781.png)

looks good!

PR #1382, opened 2022-08-16
Hotfix verification duplicate address check

RamRamez:
Ref. #1181 

1. fix duplicate address check
2. show pointer cursor on select items


MohammadPCh (approved):
Thanks @RamRamez

PR #1393, opened 2022-08-18
Discontinued GIVfarm cards

mateodaza:
Sets discontinuation for production farms to stop showing on September 8th as it is exactly 3 weeks after August 18th.

Before that date we'll display them by default and tell the user to remove all liquidity from there

@laurenluz  @divine-comedian we can change the date if you disagree

Test with production env

### DIFF
--- a/src/components/homeTabs/GIVfarm.tsx
+++ b/src/components/homeTabs/GIVfarm.tsx
@@ -54,13 +54,7 @@ const renderPools = (
 ) => {
 	const rightNow = getNowUnixMS();
 	return pools
-		.filter(p =>
-			showArchivedPools
-				? true
-				: p.active &&
-				  (!p?.discontinued || rightNow < p?.discontinued) && // shows card if time of discontinuation isn't over
-				  !p.archived,
-		)
+		.filter(p => (showArchivedPools ? true : p.active && !p.archived))
 		.map((poolStakingConfig, idx) => ({ poolStakingConfig, idx }))
 		.sort(
 			(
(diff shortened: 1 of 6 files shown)


mateodaza:
100% GIV staking on mainnet
GIV / DAI uniswap on mainnet
GIV/ETH sushiswap on gnosis chain
GIV / xDAI honeyswap on gnosis chain

This is the message
image

src/components/homeTabs/GIVfarm.tsx
```diff
@@ -145,6 +153,17 @@ export const TabGIVfarmBottom = () => {
...
+	const mainGIVStakeDiscontinued =
+		mainnetGIVStaking.discontinued &&
+		getNowUnixMS() > mainnetGIVStaking.discontinued;
+
+	const xdaiGIVStaking = getGivStakingConfig(config.XDAI_CONFIG);
+
+	const xdaiGIVStakeDiscontinued =
+		xdaiGIVStaking.discontinued &&
+		getNowUnixMS() > xdaiGIVStaking.discontinued;
+
```
MohammadPCh, line 166: please move GIVstaking configs to the out of the component, I think the networkProvider.ts is a good place for it.

src/types/config.ts
```diff
@@ -4,6 +4,8 @@ export interface BasicStakingConfig {
 	BUY_LINK?: string;
 	farmStartTimeMS?: number;
 	icon?: string;
+	active?: boolean;
```
MohammadPCh, line 7: may we should decide between this and the active property in the simpleStakingConfig,
they are same!

MohammadPCh (changes requested):
@mateodaza Thanks,
I suggest so  improvements also I have written 2 more suggestions here:

- I think we don't need to add complexity to our app if we want to archive some farms at a specific time, we can easily change them to archived at that time.
- please add configs in development configs too. it helps testers to test it easier.

thanks.

mateodaza:
As soon as we merge this, I'll create a new PR for manually archiving in Sept 8th

MohammadPCh (approved):
Thanks @mateodaza

PR #1399, opened 2022-08-19
Every GIVfarm cards shown in same view for any network

mateodaza:
Leaving the sorting feature for another [issue](https://github.com/giveth/giveth-dapps-v2/issues/1400) - Also, when the Discontinued GIVFarm card is merged we'll have a conflict to solve here


mateodaza:
replaed by [this one](#1403 )

PR #1418, opened 2022-08-21
Some tiny fixes

RamRamez:
1. add gray bar in mobile view to project card in donate page
2. fix font weight of join page titles
3. fix font size of checkboxes


alireza-sharifpour (approved):
Thanks @RamRamez

Issue #1439, opened 2022-08-28
GIVpower: showing Rank on Project Card


MoeNick:
https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=12568%3A235339

MoeNick:
![image.png](https://images.zenhubusercontent.com/604f242c2819d5da6426ad03/4018101d-514c-47fd-9aa1-dee306e5895c)
The placement of raised amount has to change and and it needs animations, so in this case it may caused lags in loading, is it okay @mosaeedi ?

mosaeedi:
It's a mistake
We should always show both, ranking and raised amount on default and hover state.

maryjaf:
I did not see any change in the project card on giveconomy compared to before @alireza-sharifpour

alireza-sharifpour:
I couldn’t understand the problem, you can see rankings on:
https://givpower.vercel.app/
@maryjaf

MoeNick:
It was an old issue @maryjaf will close it to avoid confusion.

PR #1444, opened 2022-08-30
Added givpower to porject cards

alireza-sharifpour:
- #1439


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #1437, opened 2022-08-28
showing the complete profile toast wrongly


MohammadPCh:
when the user's token is expired, and the user goes to the account page, we show a toast that says please complete your profile.

the main problem is when the user's token is expired the backend doesn't return the email, so we think the profile is not complete.

I talked with Mohammad (Giveth/impact-graph#602) and ask him to add `isSignedIn` to the `getUserByAddress` query. so we can distinguish that the user is not signed in or the profile is not complete.

maryjaf:
On which test environment can it be tested? @alireza-sharifpour

alireza-sharifpour:
@maryjaf 

You can test it on staging
https://staging.giveth.io/

maryjaf:
I tested the scenario described in the issue (when the user's token is expired, and the user goes to the account page, we show a toast that says please complete your profile.)

on the staging env, the toast related to completing the profile is not displayed correctly.
But in this scenario, like in production, the list of projects and donations, etc., is not displayed
Is this how it should be? @alireza-sharifpour 
<img width="1438" alt="image" src="https://user-images.githubusercontent.com/111529185/191933428-a467d730-d04a-4837-b540-0cb3f964c42b.png">

alireza-sharifpour:
> I tested the scenario described in the issue (when the user's token is expired, and the user goes to the account page, we show a toast that says please complete your profile.)

When the token is expired in this page the user should be signed out and see the sing in message, not the account page.

Issue #1436, opened 2022-08-28
logout with expired token


MohammadPCh:
Nothing happens when the user token is expired and the user tries to log out.

maryjaf:
How long is the lifetime of the token, and with this issue, what do we expect to happen after the token expires?
There was no figma on the task
@MohammadPCh @alireza-sharifpour

alireza-sharifpour:
> How long is the lifetime of the token, and with this issue, what do we expect to happen after the token expires? There was no figma on the task @MohammadPCh @alireza-sharifpour

Backend Guys know the exact Expiration time @mohammadranjbarz @CarlosQ96 
the issue was when we wanted to logout with invalid or expired token, the logout function didn’t work, because the token was not valid, so we couldn’t sign the user out.
know the issue is fixed.
for testing you can sign in and then change the token in localStorage or remove one or two letter from the end of it to make it invalid, after that you should be able to logout from the header menu.
@maryjaf

maryjaf:
I tested this scenario in staging and it passed
sign in
change and invalid my token from inspect-local storage
go to my account page
tap on sign out 
and i signed out correctly

PR #1453, opened 2022-09-01
Fix/user is signed in logout issues

alireza-sharifpour:
- #1436 
- #1437


MohammadPCh (approved):
Thanks @alireza-sharifpour

PR #1455, opened 2022-09-01
update projectcard design for givpower

alireza-sharifpour:
- #1439


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #1122, opened 2022-07-14
Categories & subcategories


MoeNick:
@mosaeedi commented on [Tue Jul 12 2022](https://github.com/Giveth/GIVeconomy/issues/657)

Module structure
![Screen Shot 2022-07-12 at 12.17.07 PM.png](https://images.zenhubusercontent.com/61dd85b9a1419a0fc83f15d1/fa3a25fa-de8c-43d9-9c0d-2e8c1ba1b796)

---

## Exception
**All** category:
For the **All** category, we don't have any sub-categories and the sub-categories row should be disabled and not visible

![Screen Shot 2022-07-12 at 11.51.33 AM.png](https://images.zenhubusercontent.com/61dd85b9a1419a0fc83f15d1/fa0b284c-43c1-427f-aaf0-cc1558d75e40)

---

## Design resource
**All** category selected - > [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=11080%3A201383)

Main category selected - Sub categories are visible -> [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=8026%3A172569)

---

divine-comedian:
how can this issue be tested?

alireza-sharifpour:
> how can this issue be tested?

You can test it here:
https://giveth-dapps-v2-git-categories-develop-branch-givethio.vercel.app/

MoeNick:
Im also confused, recently @MohammadPCh send me this: https://givfilter.vercel.app/

MohammadPCh:
@divine-comedian @MoeNick these are the same version of the app. I only added a new URL for easier access.

divine-comedian:
testing on `https://giveth-dapps-v2-git-categories-develop-branch-givethio.vercel.app/`

checking out categories and sub-categories 

the first thing I notice is that when you scroll sideways through the categories the `All` category disappears, it wasn't clear the functionality from the figma but I think as a user we should always be able to see All and the other categories can scroll sideways. 

The category pages also load sort of slow and it feels like the switch between categories is very clunky, you can see the banner blink and reload and same with the project cards as they reload and shuffle around in the bottom. Would adding in some sort of transition effect help this appear smoother? I made a loom just to show what currently happens 

I tried to show a bit further what I mean with the `All` category and slow category loading in a quick loom video:
[...]

PR #1475, opened 2022-09-06
Add/categories sorting and count

alireza-sharifpour:
- #1122


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #761, opened 2022-05-19
GIVpower - Single Project View - Boosting Modal


MoeNick:
- No GIVpower modal -> [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=12568%3A236170)

- Boosting modal -> [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=12568%3A232403)

- Confirmation modal -> [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=12568%3A236144)

- First time boosting.

MoeNick:
Should we remove Unallocated GIVpower @mosaeedi ?
Copy on toast for first time boosting.

MoeNick:
@divine-comedian @maryjaf pls confirm and merge.

divine-comedian:
In this view here- when a project has not been boosted yet the copy seems a bit off.

This is not my project so it should not refer to this as 'your project' copy should be:

'This project hasn't received any boosts yet!'
'Share this project on social media and ask your friends for a boost!'

- it would be cool if 'Share this project' opened the modal similar to the share button on the card on the right under the donate button.

![image](https://user-images.githubusercontent.com/67759413/190466210-f4784d4d-00e3-4cfe-b520-df505d9f6d59.png)

divine-comedian:
this project has no boosts but it is rank `#8`, how can it have a ranking with no GIVpower? 

![image](https://user-images.githubusercontent.com/67759413/190470727-245017d2-b395-41d3-a8f8-e1f9b8e55d58.png)

maryjaf:
Please check this scenario:
I'm on a GIVpower tab of a project
I select BOOST and complete the process
I expect that the boosting list will be updated at the end of the process
But this does not happen and I have to refresh the page to see the changes

And one more thing, when a project is boosted by several users, The TOTAL GIVPOWER & PROJECT RANK is displayed as pic below:
![image.png](https://images.zenhubusercontent.com/62fe7a8b57c1a1770575ea40/29854959-5bc0-4caa-82ff-5255bba00c4e)

aminlatifi:
> this project has no boosts but it is rank `#8`, how can it have a ranking with no GIVpower?
> 
[...]

PR #1478, opened 2022-09-07
761 givpower boosting functionality

alireza-sharifpour:
- #761


MohammadPCh (approved):
Thanks @alireza-sharifpour

PR #1481, opened 2022-09-07
Fix/subgraph is loaded

aminlatifi:
Moved isLoaded to subgraph value. 
There was an issue with isLoaded value, it was true but only one of subgraphs was loaded (only mainnet or only xdai)
It causes issue with calculating APR

I forked from the main branch, so it can be merged to main too. Also, I see some changes from main branch had not been moved to develop


MohammadPCh (approved):
Thanks @aminlatifi

Issue #1079, opened 2022-07-06
Add project not available page and loader to edit project


RamRamez:
1. add a loader to edit project
2. first check if user is signed in then check if user is registered. If user is not signed in, complete profile modal shouldn't be shown.
3. show project not available page in edit project if project not found
4. auto close welcome modal when user singed in, in edit project view

MoeNick:
@divine-comedian

divine-comedian:
1. loader shows successfully on edit project
2. this isn't clear on what page I should be on @RamRamez  - the homepage or edit project view?
3. I successfully see the error page visiting a project id that doesn't exist

<img width="1440" alt="image" src="https://user-images.githubusercontent.com/67759413/184925210-bbd10b7f-8496-4f47-9f1e-c3e0607f5c4f.png">

4. sign-in modal automatically closes when I sign in with metamask. this looks good!

### one small bug I found
if I am not signed in and I go to the edit project URL I see the sign in modal, if I close the sign in modal I see the infinite loading wheel for edit project
https://www.loom.com/share/be331fe75c9b47ff83c7de2c77435efd

user should be informed they need to sign in or return to homepage

mosaeedi:
Thanks for raising the issue!

Here is the design for the not signed-in state.

[Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=13439%3A242788)

RamRamez:
> Thanks for raising the issue!
> 
> Here is the design for the not signed-in state.
> 
> [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=13439%3A242788)
> 
> And the design for not connected wallet state! [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=13439%3A243454)

This deign is implemented in PR #1482 

For testing:
I added `wallet not connected` and `user not signed in` design to:
1. Verification
2. Onboarding
3. create and edit project
4. account view

I found a bug and fixed it in this PR
In some pages (like ) there is no footer. the bug is when user return to another page (that has footer) from that page, there's no footer in the new page. I fixed it in these pages:
1. Verification
5. Verification - email Verification page
6. Create and edit project page
7. success page (after create publishing a project)

Also loading component is improved in edit and create project pages.

[...]

PR #1482, opened 2022-09-07
Implement user not signed in design

RamRamez:
[#1079 ](https://github.com/Giveth/giveth-dapps-v2/issues/1079#issuecomment-1237341230)

### DIFF
--- a/src/components/views/EditIndex.tsx
+++ b/src/components/views/EditIndex.tsx
@@ -5,11 +5,7 @@ import { captureException } from '@sentry/nextjs';
 import { client } from '@/apollo/apolloClient';
 import { FETCH_PROJECT_BY_ID } from '@/apollo/gql/gqlProjects';
 import { IProjectEdition } from '@/apollo/types/types';
-import {
-	compareAddresses,
-	isUserRegistered,
-	showToastError,
-} from '@/lib/helpers';
+import { isUserRegistered, showToastError } from '@/lib/helpers';
 import CreateProject from '@/components/views/create/CreateProject';
 import { useAppDispatch, useAppSelector } from '@/features/hooks';
 import {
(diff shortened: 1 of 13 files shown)


src/components/views/create/CreateIndex.tsx
```diff
@@ -31,7 +36,16 @@ const CreateIndex = () => {
 		}
```
mateodaza, line 36: Just like in the Edit view, with this flow I believe we can remove the automatic welcome modal and use this new design, is more telling for the user instead of just forcing a login

src/components/views/EditIndex.tsx
```diff
@@ -77,9 +71,18 @@ const EditIndex = () => {
 		}
```
mateodaza, line 71: There's a flickering of the welcome modal when you are not logged and access to the route directly (try refreshing) With this new design I believe we can remove the whole modal

mateodaza (changes requested):
Besides the missing Icon that's waiting for this [PR](https://github.com/Giveth/ui-design-system/pull/70) ... just a couple of comments. besides that, excellent work @RamRamez !

src/components/views/create/CreateIndex.tsx
RamRamez: I talked with @MoeNick and he said opening welcome and sign in modal automatically is in our design flow and we can't change it.

src/components/views/EditIndex.tsx
RamRamez: like the above if we want to change the flow we should have design team approval first.

mateodaza (changes requested):
@RamRamez I'm still getting the flickering when you access the /create route directly (or refreshing it) while being not logged in

Screen-Recording-2022-09-09-at-11 39 06-AM

mateodaza (approved):

PR #1484, opened 2022-09-08
handled when the boosting project has already boosted

alireza-sharifpour:
- #761


MohammadPCh (approved):
Thanks @alireza-sharifpour

PR #1486, opened 2022-09-09
Verification revoke status check added

mateodaza:


src/apollo/gql/gqlUser.ts
```diff
@@ -50,6 +50,7 @@ export const FETCH_USER_PROJECTS = gql`
 				totalDonations
 				totalReactions
 				verified
+				verificationStatus
```
RamRamez, line 53: Is `verificationStatus` on `projectsByUserId` used anywhere?

src/apollo/types/types.ts
```diff
@@ -18,6 +18,7 @@ export interface IProject {
 	impactLocation?: string;
 	qualityScore?: number;
 	verified?: boolean;
+	verificationStatus?: string;
```
RamRamez, line 21: I think you can change `string` with `EProjectVerificationStatus`

RamRamez (changes requested):
Thanks @mateodaza, great. Just a minor fix.

src/apollo/gql/gqlUser.ts
mateodaza: Not at the moment but it's still an important property of the schema. You want me to remove it?

RamRamez (approved):
Thanks @mateodaza

PR #1488, opened 2022-09-11
1121/fix change main category issue

alireza-sharifpour:
The problem is when we are changing the mainCategory we are not removing subcategory from the state

- #1121


RamRamez (approved):
Thanks @alireza-sharifpour

PR #1506, opened 2022-09-14
Resume verification button hidden on revoked projects

mateodaza:


src/components/views/project/projectDonateCard/ProjectDonateCard.tsx
```diff
@@ -283,14 +282,16 @@ const ProjectDonateCard: FC<IProjectDonateCard> = ({
...
-						{isVerDraft && (
-							<ExternalLink href={slugToVerification(slug)}>
-								<FullOutlineButton
-									buttonType='primary'
-									label='RESUME VERIFICATION'
-								/>
-							</ExternalLink>
-						)}
+						{verificationStatus !==
+							EProjectVerificationStatus.REVOKED &&
```
RamRamez, line 286: @CarlosQ96 said `verificationStatus` will be changed with `verified` status at the same time with cronjob. Means if project verification is revoked, `verified === false` also. So I think you don't need `verificationStatus` for this task anymore.
For showing `VERIFY YOUR PROJECT` button, I think the `verified` condition is enough and you can remove `verificationStatus` condition.
And for showing `RESUME VERIFICATION` button, I think replacing `verificationStatus !== EProjectVerificationStatus.REVOKED` condition by `verified` will work.

RamRamez (changes requested):
Thanks @mateodaza. Your solution is working well. I have a suggestion please take a look.

src/components/views/project/projectDonateCard/ProjectDonateCard.tsx
mateodaza: There are two cases 
- A new project that has never been verified, the property is shown as false but we want to display the button
- A previously verified project which was revoked. We don't want them to see the verification button

If I don't check both values the first case will never see the option

RamRamez: Is it necessary to check `verificationStatus` for showing `RESUME VERIFICATION` button? I mean the project is not verified yet and verificationStatus can't be revoked.

RamRamez (approved):

PR #1511, opened 2022-09-15
text and styling imrpovements updates

divine-comedian:
fixing some copy


divine-comedian:
@MohammadPCh any idea why this fails? seems to be a node package error on build.

MohammadPCh (approved):
Thanks @divine-comedian

PR #1521, opened 2022-09-19
761 added to fixed to table amount

alireza-sharifpour:
- #761


MohammadPCh (approved):
Thanks @alireza-sharifpour

PR #1522, opened 2022-09-19
1122 replace subcategories value with name

alireza-sharifpour:
- #1122


MohammadPCh (approved):
thanks @alireza-sharifpour

PR #1527, opened 2022-09-20
Adds scroll to menu when height is low

mateodaza:


MohammadPCh (changes requested):
@mateodaza Thanks,
please look at this:
image
I suggest to don't showing the scrollbar when it's not needed.
what's your idea about using the scroll bar library that we have in our project?

Issue #1537, opened 2022-09-22
Conflict between Cypress and Jest


alireza-sharifpour:
There is a conflict between jest and cypress typescript

![Screen Shot 2022-09-22 at 1 33 04 PM](https://user-images.githubusercontent.com/48887817/191718829-764dc018-e943-4b2b-b68b-a8feaefc047d.png)

So we need to tell our main typescript compiler to limit itself to src and pages folder

alireza-sharifpour:
The solution for us would be something like this:
https://github.com/neiltownsley/react-typescript-cypress/pull/1

MohammadPCh:
@alireza-sharifpour Thanks,

maryjaf:
Should this issue be tested?
I did not understand from the description  what effect the changes can have on end-to-end scenarios
@alireza-sharifpour

alireza-sharifpour:
> Should this issue be tested? I did not understand from the description what effect the changes can have on end-to-end scenarios @alireza-sharifpour

No, this doesn’t need test. it's an internal thing.

PR #1538, opened 2022-09-22
Resolved cypress and jest conflict

alireza-sharifpour:
- #1537


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #1545, opened 2022-09-26
 Modify the start and ends dates on GIVbacks page for Tuesdays


laurenluz:
GIVbacks round 20 should run from September 16 to October 4

![image](https://user-images.githubusercontent.com/75490971/192306519-8a90f2b5-03d1-44c7-98b2-364b5e13ed70.png)

GIVbacks round 21 should run from October 4 to October 18

aminlatifi:
#1556 is merged, after round 20 finishes we must create another PR with simple round offset in the config.

MohammadPCh:
@laurenluz it's merged, so you can test it on https://giveconomy.vercel.app/givbacks and if you approve it we can have a version on production.

MohammadPCh:
@aminlatifi Thanks for reviewing, yeah, I totally agree with you, I will fix it after this round.

laurenluz:
do we need another issue for that @MohammadPCh ? Or just keep this one?

laurenluz:
it's on production!

PR #1556, opened 2022-09-27
Change rounds time

MohammadPCh:
- #1545


aminlatifi:
The code is good, but after round 20 finishes we will create another PR and have a simple round offset value in config and apply that in calculation.

Issue #1540, opened 2022-09-22
GIVpower page modifications for Staking/Locking


aminlatifi:
@laurenluz @MoeNick 
I have upgraded the Garden unipool contract corresponding to the develop env. Now it supports locking functionalities.
The start time equals token distro start time and the duration is 14 days.

I created the subgraph based on this new configuration, it's been deployed here https://thegraph.com/hosted-service/subgraph/aminlatifi/giveconomy-xdai-deployment-seven?selected=logs

What remain:
- [ ] Merge givpower to giveconomy branch and configure it to use the above subgraph temporarily. We can use develop default subgraph after this [PR](https://github.com/Giveth/giveconomy-subgraph/pull/30/files) is merged @MohammadPCh 
- [ ] Set up an unlock bot for develop env @aminlatifi

MoeNick:
This is the preview before merging to giveconomy branch
https://giveth-dapps-v2-git-givpower-ready-to-merge-givethio.vercel.app/

MoeNick:
@maryjaf this one also reported by @divine-comedian 
https://github.com/Giveth/giveth-dapps-v2/issues/1512

divine-comedian:
on the GIVpower page did we decide to remove the greyed out boost projects?
https://giveth-dapps-v2-git-givpower-ready-to-merge-givethio.vercel.app/givpower

![image](https://user-images.githubusercontent.com/67759413/192383645-f43dc6ae-9b03-4dd1-bece-fe65d37ef690.png)

If so I would suggest adding some padding to the bottom of the element, otherwise it looks a bit squished. 
![image](https://user-images.githubusercontent.com/67759413/192383811-534f29ee-3936-4446-96d2-071314468053.png)

divine-comedian:
on this page the two `?`s are aligned differently - one is centered and the other is top-aligned

another thing - the table doesn't seem to fit on one page - I have to scroll side ways to see the information. on the figma spec this should fit onto one page. 


![image](https://user-images.githubusercontent.com/67759413/192385416-5cb08665-3946-4482-81ca-8e822bf03b63.png)


https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=10527%3A194487

RamRamez:
> w, without scroll

Iremoved

laurenluz:
[...]

PR #1568, opened 2022-09-27
fix your givpower card and copy

RamRamez:
#1540


mateodaza (changes requested):
image

Excellent work @RamRamez thanks a lot, just a tiny change. Can we align the bottom texts? "Coming Soon" and "Stake GIV"

Issue #1599, opened 2022-10-03
Prevent user to boost more then 20 projects from single project view


MoeNick:
Error msg should be shown on a boost modal while user have boosted the 20 project before:

You have already boosted the maximum 20 projects. 
Do you want to boost this project anyway? Try to remove at least one other boosted project from your account, and get back to this project again!

laurenluz:
Copy fix:

"You have already boosted the maximum 20 projects!
To coninute with this boosting, remove at least one other boosted project from your account and come back to this project again!"

Button: "Go to Boosted Projects"

mosaeedi:
![Screen Shot 2022-10-03 at 6 07 32 PM](https://user-images.githubusercontent.com/13708322/193611659-a15f110e-6dec-4cec-923b-cab69361cf94.png)


[Link to Figma](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=12568%3A236170)

MoeNick:
Sorry @alireza-sharifpour fr the changes, can you jump into it again?

maryjaf:
I think the changes are not deploy on https://givpower.vercel.app/
No change was observed in the scenario of the upper limit of 20 @alireza-sharifpour @MohammadPCh

maryjaf:
Thanks guys, this is ok.
![image.png](https://images.zenhubusercontent.com/62fe7a8b57c1a1770575ea40/b6b42632-0021-4418-ad7d-0a6effa32abb)

PR #1600, opened 2022-10-03
handle max 20 boosted projects

alireza-sharifpour:
- #1599


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #758, opened 2022-05-19
GIVpower | My Account - Boosted Projects-  Manage GivPower


MoeNick:
- Edit boosting -> [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=12568%3A236357)

- Edit boosting (amounts changed) -> [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=12568%3A238961)

- update boosting confirmation modal -> [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=13185%3A240957)

- Removing a boost modal -> [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=12771%3A233321)

- Removing the last boosted project error modal -> [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=12771%3A233434)

**Edge cases**
- only 3 boost and locking -> [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=13029%3A251757)
- exceeding 100% -> [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=13046%3A253456)

## Guides 

- Responsive guides -> [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=12921%3A236174)

- Mobile design -> [Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=12921%3A236339)

MoeNick:
@divine-comedian @maryjaf pls confirm and merge.

maryjaf:
I didn't see this modal when i applied some change in boosting percentage

![image.png](https://images.zenhubusercontent.com/62fe7a8b57c1a1770575ea40/a9a18e69-8233-4817-814a-55ff1c0471e3)

MohammadPCh:
@maryjaf Thanks for reminding me. I will leave a PR for it today.

[...]

PR #1609, opened 2022-10-04
Some fixes on boosts table

MohammadPCh:
fix these issues:
- problem on edit when the user only has one boost (#758)
- fix boost table performance
- handle sort boosts on front-end


alireza-sharifpour (approved):
Thanks @MohammadPCh

Issue #1558, opened 2022-09-27
Frontend: Adding "Accept Funds on Gnosis" filter to givfilter


MoeNick:

MoeNick:
@alireza-sharifpour 
@mohammadranjbarz has just updated the backedn, I assumed it;s ready for test, if it;s not pls let us know.

alireza-sharifpour:
> @alireza-sharifpour @mohammadranjbarz has just updated the backedn, I assumed it;s ready for test, if it;s not pls let us know.

Yes, it can be tested on:
https://givfilter.vercel.app/projects
@MoeNick @maryjaf

PR #1619, opened 2022-10-05
added Accept Funds on Gnosis

alireza-sharifpour:
- #1558


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #1559, opened 2022-09-27
SEO - include text in the body under the main banner. 


MoeNick:
originally from this comment: https://github.com/Giveth/giveth-planning/issues/865#issuecomment-1252435325

Can you add a section for text below the image, so we can have content for SEO purposes (just one small paragraph ex.
"Donate crypto directly to Economics and Infrastructure projects on Giveth! Support projects and get rewarded for giving with zero added fees."

- [ ] Design amendments by @mosaeedi 

Text suggestions by @ahmadabugosh under these main categories are required:

- [x] Art & Culture
- [x] Community
- [x] Economic and Infrastructure
- [x] Education
- [x] Equality
- [x] Finance
- [x] Health & Wellness
- [x] Nature
- [x] NGO
- [x] Technology

MoeNick:
wow! so quick, thanks! @ahmadabugosh

mosaeedi:
![Screen Shot 2022-09-28 at 3 45 36 PM](https://user-images.githubusercontent.com/13708322/192781917-ed0d6de0-8bdb-4494-8576-df4fd4d2553e.png)

mosaeedi:
@ahmadabugosh 
I added a new section between the project cards, the "Zero added fees!" white section, you can have a title and a subheading to add anything you want.

Also, did you notice there is a subheading under the main category title? below the "Environment & energy" it's a dynamic text holder that you can use for each main category!

mosaeedi:
Here is the link to Figma file -> [link to Figma](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=13866%3A248323)

mosaeedi:
cc @MoeNick

mosaeedi:
The white row, between the product, should appear just after the first row!

MoeNick:
What are the use-cases in filtering or sorting? this shows only on main category filters? 
what if the main category don't have enough projects? one row for example? 
@mosaeedi

ahmadabugosh:
Hey @mosaeedi thanks for this! 

A few points to follow up on:

1) Would the text I sent for the description fit on the top header image as a subtitle? If not you suggest do you suggest we change it to something else? If it does fit, maybe we don't need the white row at all. 

[...]

PR #1638, opened 2022-10-10
1559 add middle banner

alireza-sharifpour:
- #1559


RamRamez (approved):
Thanks @alireza-sharifpour

Issue #1676, opened 2022-10-19
GIVpower Boosting | Tooltip on Boosting modal - text fix


laurenluz:
**Describe the bug**
Tooltip grammar is off... should say "This is the number of projects you have boosted before."

**To Reproduce**
1. Go to Project Page
2. Click on 'Boost'
3. Mouse over '?"
4. See tooltip

**Expected behavior**
Should say "This is the number of projects you have boosted before."

**Screenshots**
![image](https://user-images.githubusercontent.com/75490971/196701111-9c37b7c0-b0b0-4af2-a063-6adc2d921c53.png)

**This happens on which Boosting Section?**
- [ ] Project's Card - GIVpower Rank
- [ ] Project's GIVpower ranking
- [ ] Single Project Public View | Boost Button
- [ ]  Single Project's owner view | Boost Button
- [ ]  Single Project's View | Givpower tab (Hall of fame, Total Boosted) 
- [x] Boost Modal for a specific project
- [ ] Boost Succeeded
- [ ] My Account | Boosting Tab: list of boosted projects
- [ ] My Account | Delete a Boosted Projects
- [ ] My Account | Edit boosting allocations
- [ ] My Account | Success editing allocations
- [ ] My Account | Success editing allocations.

Issue #1673, opened 2022-10-18
GIVpower Boosting | not able to boost my own project


NikolaCreatrix:
**Describe the bug**
Using google chrome. Signed in, giv staked and locked, on my own projects page there is no option to boost my own project.

![image](https://user-images.githubusercontent.com/100084867/196514785-f1aee66e-d951-413b-9c2c-a9b243931111.png)

![image](https://user-images.githubusercontent.com/100084867/196515065-e5d9eb2c-ce3f-4fa1-b183-a42a97513b00.png)

**To Reproduce**
1. Go to 'my account after sign in'
2. Click on 'view' to one of your active, listed projects

4. See error - no option to boost your own project

**Expected behavior**
add the 'boost' button

**This happens on which Boosting Section?**

- [ ]  Single Project's owner view | Boost Button

laurenluz:
@MohammadPCh can you take this?

laurenluz:
ooh @NikolaCreatrix boosting should only be possible on already verified projects (not ones in the middle of the verification process). Did you test with an already verified project?

NikolaCreatrix:
Hi @laurenluz and @MohammadPCh 

I just tried with this project

https://giveconomy.vercel.app/project/giveth-tour-bridge-the-world

![image](https://user-images.githubusercontent.com/100084867/196745823-55aadc3d-555e-4daf-ba22-d876b4292d21.png)

Still no Boost button.

MohammadPCh:
@NikolaCreatrix it's merged about 20 minutes ago, can you check it again?

MoeNick:
@NikolaCreatrix based on recent changes you need to have a verified project to be able to boost it, is your project verified?

NikolaCreatrix:
Hey @MoeNick and @MohammadPCh 

Thanks, it works now!

![image](https://user-images.githubusercontent.com/100084867/197028969-1aaa5a66-a31d-4d00-94a3-db11f532697e.png)

Issue #1670, opened 2022-10-18
GIVpower Boosting - Why is there no number for my boosted projects


divine-comedian:
**Describe the bug**
In every other tab you can see a total of items that correspond to the tab heading

i.e how many projects i own, how many donations I've made, how many projects I've liked. Why is there no similar functionality for boosted projects? 

**To Reproduce**
1. log into giveth
2. Click on my account

**Expected behavior**
A user should be able to see a small number next to `boosted projects` representing the number of projects they have boosted.

**Screenshots**
![image.png](https://images.zenhubusercontent.com/5fff1c86d883b3d9a858ddbc/276e2249-49c2-4d1a-88c7-c3b070438e9a)

**This happens on which Boosting Section?**
- [ ] Project's Card - GIVpower Rank
- [ ] Project's GIVpower ranking
- [ ] Single Project Public View | Boost Button
- [ ]  Single Project's owner view | Boost Button
- [ ]  Single Project's View | Givpower tab (Hall of fame, Total Boosted) 
- [ ] Boost Modal for a specific project
- [ ] Boost Succeeded
- [x] My Account | Boosting Tab: list of boosted projects
- [ ] My Account | Delete a Boosted Projects
- [ ] My Account | Edit boosting allocations
- [ ] My Account | Success editing allocations
- [ ] My Account | Success editing allocations. 

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.05
 - Browser firefox

laurenluz:
It should have a number for sure - @MohammadPCh can you take it?

Issue #1669, opened 2022-10-18
GIVpower Boosting - GIVpower page styling and missing link


divine-comedian:
**Describe the bug**
on the GIVpower page
https://giveconomy.vercel.app/givpower

the button learn more is way too big 

also where it says coming soon we should add the link to allow users to boost projects and change the text
**To Reproduce**
1. Go to https://giveconomy.vercel.app/givpower

**Screenshots**
![image](https://user-images.githubusercontent.com/67759413/196481119-2462d4d1-35cf-492c-9465-3aa2b4468850.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04
 - Browser firefox

laurenluz:
The learn more button is tested & looking right on staging now.

The change to "coming soon" should be done in #1698

moving to done!

laurenluz:
![image](https://user-images.githubusercontent.com/75490971/197618577-4fc864d1-7bc2-4cdc-8939-18525a21a3d9.png)

PR #1680, opened 2022-10-19
Fix testparty issues

MohammadPCh:
fix these issues:
- #1669 
- #1670 
- #1676 
- #1673 
- remove GIVpower on the not verified projects tab from the single project page
- show GIVpower rank in project card only for verified projects.


alireza-sharifpour (approved):
Thanks @MohammadPCh

PR #1683, opened 2022-10-20
Fix boosted projects counter

alireza-sharifpour:
fixed boostTab counter issues when we are deleting a boosted project


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #1675, opened 2022-10-19
Similar projects are not displayed 


maryjaf:
By selecting a project, the list of similar projects is empty in UI, but in the response of graphql, a list of similar projects is displayed.
<img width="1430" alt="image" src="https://user-images.githubusercontent.com/111529185/196698787-18b4c67e-c9bd-4697-a7b4-f36c56d0f643.png">

MohammadPCh:
@RamRamez, can you take it please? Thanks :)

PR #1685, opened 2022-10-20
1675 fix similar projects issue

alireza-sharifpour:
- #1675


RamRamez (approved):
Thanks @alireza-sharifpour

Issue #1692, opened 2022-10-21
Filters drop down text fix


laurenluz:
Should really say "Accept GIV" not "Accept GIV token"

![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/ef9580f3-96f1-43ca-9972-e1ce97e6407a)

PR #1705, opened 2022-10-26
change Accept GIV Token to Accept GIV

alireza-sharifpour:
- #1692


RamRamez (approved):
Thanks @alireza-sharifpour

PR #1724, opened 2022-10-31
Check isSignedIn and check for new notifications

alireza-sharifpour:


MohammadPCh (approved):
LGTM, thanks @alireza-sharifpour

Issue #1742, opened 2022-11-02
Remove inactive partnership's from Giveth website


MoeNick:
@SnakeGH commented on [Wed Nov 02 2022](https://github.com/Giveth/giveth-dapps-v2/issues/1751)

After speaking with Griff, he mentioned that these two partnerships are inactive, and Giveth will not be working with them moving forward. They can be removed from our partnership page on the Giveth website

-District0x
-SourceCred

@MoeNick would you be able to assist with this? 

---

@MoeNick commented on [Wed Nov 02 2022](https://github.com/Giveth/giveth-dapps-v2/issues/1751)

sure! Maybe @alireza-sharifpour can fix it.

divine-comedian:
testing on https://giveth.io/partnerships

I can no longer find the partnerships listed above on the page! 

this looks to be live already on production so I will close this issue

PR #1743, opened 2022-11-02
1742 remove partnerships

alireza-sharifpour:
- #1742


MohammadPCh (approved):
Thanks @alireza-sharifpour

PR #1778, opened 2022-11-08
fix icons import export

alireza-sharifpour:


MohammadPCh (approved):
@alireza-sharifpour Thanks,

PR #1779, opened 2022-11-08
fix icon notification issue

alireza-sharifpour:


MohammadPCh (approved):
Thanks @alireza-sharifpour

PR #1791, opened 2022-11-10
Fix notifications styling

alireza-sharifpour:
fixed styling issues


MohammadPCh (approved):
Thanks @alireza-sharifpour

PR #1814, opened 2022-11-14
Fix tooltip position

alireza-sharifpour:
fixed tooltips sizing on mobile


MohammadPCh (approved):
@alireza-sharifpour Thanks, LGTM.

Issue #1832, opened 2022-11-16
Infinite scroll for notifications page


alireza-sharifpour:
Based on the @mosaeedi's opinion, we should implement an infinite scroll for notifications center like the one we have in projects page.

divine-comedian:
Where can this issue be tested?

alireza-sharifpour:
> Where can this issue be tested?

You can test it here @divine-comedian 
https://notification-center-13.vercel.app/

divine-comedian:
> > Where can this issue be tested?
> 
> You can test it here @divine-comedian https://notification-center-13.vercel.app/

thank you - are there any tricks to quickly generating many notifications?

alireza-sharifpour:
> > > Where can this issue be tested?
> > 
> > 
> > You can test it here @divine-comedian https://notification-center-13.vercel.app/
> 
> thank you - are there any tricks to quickly generating many notifications?

The only way I use it is to donate to projects and see the notification. It's not an optimal way!
Maybe @mohammadranjbarz knows a better way.

mohammadranjbarz:
@divine-comedian  @alireza-sharifpour 

Another way to create lots of notifications is to use admin bro and make your projected, listed, verified, unverified, activate, cancelled, ....

Another way is to create notification by calling a webservice ( it's the webservice that impact-graph use that for creating notifications) so if you wanted to use this way DM me I'll explain to you, but I think admin-bro way is easier
https://staging.notification.giveth.io/docs/#/Notification/SendNotification

divine-comedian:
> @divine-comedian @alireza-sharifpour
> 
> Another way to create lots of notifications is to use admin bro and make your projected, listed, verified, unverified, activate, cancelled, ....
> 
> Another way is to create notification by calling a webservice ( it's the webservice that impact-graph use that for creating notifications) so if you wanted to use this way DM me I'll explain to you, but I think admin-bro way is easier https://staging.notification.giveth.io/docs/#/Notification/SendNotification

I think I can do the first option pretty easily, thanks!

PR #1834, opened 2022-11-16
Handle notifications infinite scroll

alireza-sharifpour:
- #1832


MohammadPCh (approved):
@alireza-sharifpour thanks.

Issue #1855, opened 2022-11-22
GIVpower Boosting |  GIVpower boosting  in the user list and in the project list are different 


maryjaf:
**Describe the bug**
I boosted the project
At the same time, it was displayed in the project boosting list
But the givpower I see in the user list and in the project list are different

**To Reproduce**
1. Go to 'https://giveconomy.vercel.app/user/0xcd192b61a8dd586a97592555c1f5709e032f2505?tab=givpower'
2. check for example the givpower of  'KIPP Foundation' project
3. Go to 'https://giveconomy.vercel.app/project/kipp-foundation'
4. check the givpower of  this user 'maryam jafarymehr'

![image](https://user-images.githubusercontent.com/111529185/203321691-333ceaec-6d0b-42f9-b821-c68cb0fe4fd3.png)
![image](https://user-images.githubusercontent.com/111529185/203321715-1e952076-ab73-46c9-820d-4e5caa8e68d6.png)

PR #1859, opened 2022-11-22
fixed givpower value issue

alireza-sharifpour:
- #1855


MohammadPCh (approved):
Nice catch, Thanks @alireza-sharifpour

PR #1873, opened 2022-11-26
Change "Gnosis chain" name to match https://chainlist.wtf/

abitrolly:
This should fix Metamask warning about name not matching chain ID.

image

### DIFF
--- a/src/config/production.ts
+++ b/src/config/production.ts
@@ -205,7 +205,7 @@ const config: EnvConfig = {
 
 	XDAI_CONFIG: {
 		chainId: '0x64',
-		chainName: 'Gnosis Chain',
+		chainName: 'Gnosis',
 		nativeCurrency: {
 			name: 'XDAI',
 			symbol: 'XDAI',


RamRamez:
This issue is addressed in #2296 
@abitrolly You can close it

abitrolly:
@RamRamez I don't think it is a good idea to merge Gnosis related change in a Polygon related PR. My commit is also easier to review and merge (and came first), so I will keep it open for now. Thanks for letting me know, though.

RamRamez:
BTW, you've changed the wrong file.

abitrolly:
@RamRamez it appears there are quite a lot of files with all the same data.

https://github.com/search?q=repo%3AGiveth%2Fgiveth-dapps-v2%20%22gnosis%20chain%22&type=code

Looks like there is no single source of truth. Why all these are needed?

abitrolly:
Looks like its a tough competition to get into GIVeconomy. :D

PR #1903, opened 2022-12-07
fix projectcardalt link styling issue

alireza-sharifpour:


MohammadPCh (approved):
@alireza-sharifpour Thanks, I will take a look at the `InternalLink` component, and check its all usages.

Issue #1939, opened 2022-12-15
Move project data to context


alireza-sharifpour:
Currently we have a lot of prop drilling in our project section and it's better to move them to project context and read data from there.
It will improve the code readability and performance.

PR #1940, opened 2022-12-15
Move project data to context

alireza-sharifpour:
#1939


MohammadPCh (commented):
@alireza-sharifpour Thanks.

Issue #1938, opened 2022-12-15
No link on the New Project button after Creating a profile


Cotabe:
GM wonderful dev team!

There is no link on the New Project button after a person just created their profile.

![Create project after onboard bug.PNG](https://images.zenhubusercontent.com/6239faf88ad3a4ffddc12d98/ceab3bcb-2885-4201-8e37-097602563f3b)

PR #1964, opened 2022-12-22
fix new project button issue on onboarding last step

alireza-sharifpour:
#1938


MohammadPCh (approved):
@alireza-sharifpour thanks

PR #1980, opened 2022-12-28
Hostname added instead of pathname on token signature for SIWE

mateodaza:
In our current implementation the signToGetToken function will always be called on the client so we can get the host straight from the window, if for some reason in the future we call it from the server we can get the value from the req


aminlatifi (changes requested):
These changes don't solve the problem!
Anyway, we had already used window.location.hostname here
https://github.com/Giveth/giveth-dapps-v2/blob/79c6285873f2c4e08e2889281e7b36c4606422aa/src/lib/helpers.ts#L443

aminlatifi:
> In our current implementation the signToGetToken function will always be called on the client so we can get the host straight from the window, if for some reason in the future we call it from the server we can get the value from the req

The intention of issue was solving the problem for server side rendering

Issue #1953, opened 2022-12-19
Sign-out refactoring on front-end (session management) 


MoeNick:
As a user I want the dapp to recognize me if I had a successful sign-in before. (until my token is valid)

MohammadPCh:
@maryjaf could you please test this? I can explain the scenario if you don't know it.

maryjaf:
Thanks @MohammadPCh I have tested this and once the authentication is complete, like and unlike actions are performed
and the modal related to boost and donate is displayed correctly for both cancel and sign scenarios

maryjaf:
Plz check this @alireza-sharifpour 
Sometimes, even though I am not logged in, I see the sign out option in the menu
As you can see in the picture, the sign out option is displayed
But when I click like, for example, the modal sign is opened
<img width="798" alt="image" src="https://user-images.githubusercontent.com/111529185/210005410-6ad945f7-2561-41e5-a75b-b5947e5160c5.png">
<img width="1243" alt="image" src="https://user-images.githubusercontent.com/111529185/210005592-1fa45a8b-206a-438a-889f-7d6b5572a30a.png">

alireza-sharifpour:
Thanks @maryjaf for testing, I will take a look.

maryjaf:
Please check this scenario
I am signed in with  account 1
I switch from account 1 to account 2 ((I am not signed in, but I see the o "sign out" in menu)
By clicking on the "boost" option, boosting modal is shown
 by clicking on "confirm"  the modal related to authentication will be displayed
After completing the sign-in process, I will return to  Boost modal
By clicking on the "Confirm ", the boosting  operation is not completed and the loading remains
<img width="598" alt="image" src="https://user-images.githubusercontent.com/111529185/210106496-19208067-e7d2-49e5-b3a9-e0e4ab3c34e6.png">
@alireza-sharifpour

alireza-sharifpour:
Thanks again @maryjaf, I'll work on it.

alireza-sharifpour:
@maryjaf I did some changes, could you please test it again?

maryjaf:
> @maryjaf I did some changes, could you please test it again?

Thanks @alireza-sharifpour I tested again and i couldn't reproduce two above scenarios and It seems that the problem has been solved with the recent  your changes

maryjaf:
plz check this @alireza-sharifpour 
I liked for the first time one project with my account 1
when I switched to my account 2
[...]

PR #1987, opened 2022-12-31
Fix session

alireza-sharifpour:
#1953


MohammadPCh (approved):
Thanks @alireza-sharifpour

PR #2000, opened 2023-01-03
fixed z-index problem of SuccessView

alireza-sharifpour:


RamRamez (approved):
Thanks @alireza-sharifpour

Issue #2007, opened 2023-01-05
update notifications page if new notification received


MohammadPCh:
we need to update notifications page if new notification received.

MoeNick:
@maryjaf please take a look

maryjaf:
Thanks @alireza-sharifpour I checked this and it passed

PR #2008, opened 2023-01-05
Recieve notifications when new notification arrives!

alireza-sharifpour:
- #2007


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #2026, opened 2023-01-10
word-break issue notification menu


alireza-sharifpour:
<img width="290" alt="Screen Shot 2023-01-10 at 4 28 04 PM" src="https://user-images.githubusercontent.com/48887817/211557822-e5f5b9d5-03ec-4939-a3f8-dd18db5ea700.png">

MoeNick:
@alireza-sharifpour can you provide teh test scenario for @maryjaf

alireza-sharifpour:
If the project name is something like a WalletAddress(without spaces) and that project receives a like it can be tested.
@maryjaf

maryjaf:
Thanks @alireza-sharifpour It's ok
<img width="259" alt="image" src="https://user-images.githubusercontent.com/111529185/214340348-44f85832-5629-4e5a-8bf8-8f38b80799a5.png">

PR #2027, opened 2023-01-10
Word break issue notification menu

alireza-sharifpour:
#2026


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #1997, opened 2023-01-02
Intro block


MoeNick:
## Responsive guides

**(>1280)**
[https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=17955%3A231153&t=HA0MWRVrAq4gj0Hq-4](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=17955%3A231153&t=HA0MWRVrAq4gj0Hq-4)

**(>1024 - <1279)**
[https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=17955%3A233991&t=HA0MWRVrAq4gj0Hq-4](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=17955%3A233991&t=HA0MWRVrAq4gj0Hq-4)

**(>768 - <1023)**
[https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=17955%3A234602&t=HA0MWRVrAq4gj0Hq-4](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=17955%3A234602&t=HA0MWRVrAq4gj0Hq-4)

**(Mobile)**
[https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=17955%3A235217&t=HA0MWRVrAq4gj0Hq-4](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=17955%3A235217&t=HA0MWRVrAq4gj0Hq-4)

---
## Design asset

[Download the Exported illustration here](https://app.zenhub.com/files/448919340/09575db7-c961-4c7a-891e-4333a499ed19/download)

---

![Screenshot 2023-01-10 at 3.23.17 PM.png](https://images.zenhubusercontent.com/61dd85b9a1419a0fc83f15d1/362f18c9-3d47-48a3-89e5-f07ab53a09e5)

mosaeedi:
@rodricastillo9013 We need to replace the illustration on this intro section on the new Homepage

[...]

PR #2034, opened 2023-01-12
1997 intro block

alireza-sharifpour:
#1997


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #2015, opened 2023-01-09
Video Block


MoeNick:
### (>1280)
[https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=18282%3A247654&t=aiPuxojK60VosJQg-4](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=18282%3A247654&t=aiPuxojK60VosJQg-4)

### (>1024 - <1279)
[https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=18282%3A247709&t=aiPuxojK60VosJQg-4](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=18282%3A247709&t=aiPuxojK60VosJQg-4)

### (>768 - <1023)
[https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=18282%3A247755&t=aiPuxojK60VosJQg-4](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=18282%3A247755&t=aiPuxojK60VosJQg-4)

### (Mobile)
[https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=18282%3A247799&t=aiPuxojK60VosJQg-4](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=18282%3A247799&t=aiPuxojK60VosJQg-4)

maryjaf:
On which test environment can it be tested? @alireza-sharifpour

alireza-sharifpour:
> On which test environment can it be tested? @alireza-sharifpour

It can be tested here:
https://giveth-homepage.vercel.app/
@maryjaf

ahmadabugosh:
Here you go: https://drive.google.com/file/d/1nRkaHo9NIdqfm8TLCDLm577ePofogl9a/view?usp=share_link

maryjaf:
> > On which test environment can it be tested? @alireza-sharifpour
> 
> It can be tested here: https://giveth-homepage.vercel.app/ @maryjaf

[...]

PR #2042, opened 2023-01-17
2015 video block

alireza-sharifpour:
- #2015


MohammadPCh (approved):
@alireza-sharifpour Thanks

PR #2044, opened 2023-01-17
Gnosis Safe auto connection

mateodaza:
As a requirement made on this [PR](https://github.com/5afe/safe-apps-list/issues/203) we need to auto connect to gnosis safe when we're inside the environment

This PR makes that happen and tries to make a connection, if it fails it keeps the regular flow

It's important so we're listed on Gnosis Safe's app

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "givethdapp",
-	"version": "2.7.1",
+	"version": "2.7.2",
 	"private": true,
 	"scripts": {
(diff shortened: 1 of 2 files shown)


MohammadPCh:
@mateodaza Thanks, could you please increase the DApp version?

mateodaza:
@MohammadPCh done!

MohammadPCh (approved):
Thanks @mateodaza

PR #2047, opened 2023-01-19
Update notifications copies

alireza-sharifpour:


MohammadPCh (approved):
thanks @alireza-sharifpour

PR #2061, opened 2023-01-21
Updates Gitcoin banner design

mateodaza:


mateodaza:
@maryjaf please take a look, I merged directly to production as it's an urgent matter

Issue #1751, opened 2022-11-02
Remove inactive partnership's from Giveth website


SnakeGH:
After speaking with Griff, he mentioned that these two partnerships are inactive, and Giveth will not be working with them moving forward. They can be removed from our partnership page on the Giveth website

-District0x
-SourceCred

@MoeNick would you be able to assist with this?

SnakeGH:
@alireza-sharifpour & @MoeNick, I learned of one more, MyCrypto was acquired and merged with MetaMask so I believe that one is not valid to have our on our website anymore as well.

MoeNick:
@alireza-sharifpour I think we missed MyCrypto

alireza-sharifpour:
> @alireza-sharifpour I think we missed MyCrypto

I will work on it

MoeNick:
@maryjaf fyi to test and confirm

maryjaf:
I checked and mycrypto has been removed from list.

PR #2086, opened 2023-01-25
remove mycrypto partnership

alireza-sharifpour:
#1751


MohammadPCh (approved):
Thanks @alireza-sharifpour

PR #2117, opened 2023-02-02
Removes gitcoin alpha banner

mateodaza:


mateodaza:
Auto merging as it's not adding new code and it's relevant as gitcoin grant is over

Issue #2090, opened 2023-01-25
PFP | Overview page


MoeNick:

divine-comedian:
https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=14997%3A268092&t=XLNY6l9fJJV9fkxv-1

alireza-sharifpour:
I need the file for these 2 images on the edge!
When i save them from Figma they are cut in half!
@Tosinolawale @mosaeedi @MoeNick 
![4 1](https://user-images.githubusercontent.com/48887817/215709833-5b2749e7-6db5-4b96-a545-7e29d9091245.png)

Tosinolawale:
> I need the file for these 2 images on the edge!
> 
> <img alt="Screen Shot 2023-01-31 at 12 09 27 PM" width="392" src="https://user-images.githubusercontent.com/48887817/215710305-419a6b16-acbc-42a5-8bea-7f10d922f566.png">
> 
> When i save them from Figma they are cut in half! @Tosinolawale @mosaeedi @MoeNick ![4 1](https://user-images.githubusercontent.com/48887817/215709833-5b2749e7-6db5-4b96-a545-7e29d9091245.png)

https://forum.giveth.io/t/the-givers-pfp-collection-initial-sketches/656/57?u=santigs
Grab the images from this forum post

alireza-sharifpour:
In Forum post i only can download the jpg version, i need png version,
and also in Figma the pictures have box shadow in them so i need that in figma.
is there any way I can export it from Figma like other photos? @Tosinolawale

divine-comedian:
> 

I can help you get the raw files - send me a DM on discord.

PR #2135, opened 2023-02-08
Pfp overview styles

alireza-sharifpour:
- #2090


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #1993, opened 2023-01-02
Latest Donations Ticker


MoeNick:
## Note
This is part of the **Why Giveth** section and as we may use this on other screens, we also shared the separated row here as well.

To check the **Why Giveth** section please go to issue #1992

## Responsive guides

**(>1280)**
[https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=17980%3A245808&t=HA0MWRVrAq4gj0Hq-4](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=17980%3A245808&t=HA0MWRVrAq4gj0Hq-4)

**(>1024 - < 1279)**
[https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=17980%3A245829&t=HA0MWRVrAq4gj0Hq-4](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=17980%3A245829&t=HA0MWRVrAq4gj0Hq-4)

**(>768 - <1279)**
[https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=17980%3A245850&t=HA0MWRVrAq4gj0Hq-4](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=17980%3A245850&t=HA0MWRVrAq4gj0Hq-4)

**(Mobile)**
[https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=17980%3A245871&t=HA0MWRVrAq4gj0Hq-4](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=17980%3A245871&t=HA0MWRVrAq4gj0Hq-4)

---

![Screenshot 2023-01-10 at 4.04.03 PM.png](https://images.zenhubusercontent.com/61dd85b9a1419a0fc83f15d1/ff26119c-2f64-43e5-aba7-6aab7e7a7a84)

MoeNick:
if it's still not testable with staging data, let's put it back in progress and assigned a backend issue

[...]

PR #2157, opened 2023-02-10
latest donations ticker new design

RamRamez:
#1993


mateodaza (approved):
Excellent work @RamRamez looking pretty slick!

PR #2183, opened 2023-02-15
Add i18 to video block

alireza-sharifpour:


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #1995, opened 2023-01-02
Custom Banners Block


MoeNick:
As a Giveth user I want to see a list of active campaigns in new homepage, As a comms team member I want to have these active campaigns to users. 

![image.png](https://images.zenhubusercontent.com/604f242c2819d5da6426ad03/3ed589ea-d5fd-45f9-b38a-68fbd7624315)

PC Design:
https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=17711%3A350162&t=rxbw3GyI6hSF3HN0-4

MoeNick:
@mosaeedi please provide Mobile and tablet view design links to this issue

MoeNick:
@c0ric0ri fyi
please take a look at this issue and the backed issue: https://github.com/Giveth/impact-graph/issues/806

MohammadPCh:
@MoeNick Yes for sure,
In the first step, we need to separate the banners by type, I think we have 2 types of blocks:
1. Block with projects
1.1 block with selected projects
1.2 block with filtered projects
2. Block with video

we need to add each block's properties and separate them completely.
Then we can start implementign it on adminJS and Dapps simultaneously.
@mohammadranjbarz what do you think?

MoeNick:
What do you mean block by projects? There is only video in these blocks or a banner. We have another featured projects block which is dirrent from this.

maryjaf:
when I set  **Type : manually** and **"sorting filed: for example: newest "**
The display of the campaign is not correct and there is a problem
![image.png](https://images.zenhubusercontent.com/62fe7a8b57c1a1770575ea40/00e6a1be-fe67-48b6-b364-d6fc212976f6)
@mohammadranjbarz

mohammadranjbarz:
> when I set **Type : manually** and **"sorting filed: for example: newest "** The display of the campaign is not correct and there is a problem
> 
[...]

PR #2195, opened 2023-02-16
1995 campaigns block

MohammadPCh:
add this:
- #1995

And finalized the home page by these:
- fix paginations
- fix navigation
- update the projects page to handle query params
- etc...


alireza-sharifpour (approved):
Thanks @MohammadPCh

PR #2245, opened 2023-02-22
fix GIV and ETH token price fetching

RamRamez:


MohammadPCh (approved):
LGTM, Thanks @RamRamez

PR #2261, opened 2023-02-26
fix giveth-swag section

alireza-sharifpour:


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #2003, opened 2023-01-04
Show zero give power in the list of project give powers


maryjaf:
**Describe the bug**
when a user unstake your givpower and the givpower of user changed to zero , in the list of project boosts, this user is still displayed with 0 Give Power

<img width="954" alt="image" src="https://user-images.githubusercontent.com/111529185/210524092-4571dec2-dd93-4864-944b-bea2367c150b.png">

**Expected behavior**
In this situation, should the user be removed from the list? @laurenluz

laurenluz:
yes! we should remove their boost from the list in that case.

@MohammadPCh - can you fix it?

MohammadPCh:
Do we need to release a hotfix for it? or just put it on the staging? @laurenluz

laurenluz:
sorry for the delay @MohammadPCh !

@maryjaf - can you test?

maryjaf:
> sorry for the delay @MohammadPCh !
> 
> @maryjaf - can you test?

I tested this and now when I unstake my all givpower , my user is removed  from the list of project's givpower
but in my givpower It is displayed as below:
![image](https://user-images.githubusercontent.com/111529185/213406571-1ac180d9-3703-471e-ae9d-409507080ec0.png)
I think it should be removed from this list as well? Am I right?

MohammadPCh:
@maryjaf It totally depends on PM's decision :) but I think it's good to show them somewhere to the user.

MoeNick:
@laurenluz can you confirm this scenario?

laurenluz:
This is an interesting edge case, I think it is nice to "remember" the user's allocations even if they unstake their GIV, so that if they stake again, it is already distributed.

However, one 1st glance it's confusing, so I think we should have some kind of message if this happens. Something that says (above the table, maybe in red @mosaeedi ? )

"Your GIVpower balance is zero! Stake GIV to boost these projects again."

"Stake GIV" should link to the GIVfarm.

![image](https://user-images.githubusercontent.com/75490971/214151741-f96e5c6e-e3ee-4868-9808-94c5a4637260.png)

wdyt @MohammadPCh @maryjaf ?

MoeNick:
@mosaeedi @Tosinolawale please update us

MoeNick:
Hey @mosaeedi @Tosinolawale pls at east add a reaction! :)

mosaeedi:
## Design resource

[...]

PR #2276, opened 2023-02-27
show toast modal for zero GivPower

RamRamez:
#2003

### DIFF
--- a/src/components/views/userProfile/boostedTab/ProfileBoostedTab.tsx
+++ b/src/components/views/userProfile/boostedTab/ProfileBoostedTab.tsx
@@ -219,7 +219,9 @@ export const ProfileBoostedTab: FC<IUserProfileView> = ({
 					)}
 				</Col>
 			</Row>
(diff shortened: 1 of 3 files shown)


src/components/views/userProfile/boostedTab/ProfileBoostedTab.tsx
```diff
@@ -215,6 +219,19 @@ export const ProfileBoostedTab: FC<IUserProfileView> = ({
 					)}
 				</Col>
 			</Row>
+			{isZeroGivPower ? (
```
MohammadPCh, line 222:
```suggestion
			{boostedProjectsCount &&
			boostedProjectsCount > 0 &&
			isZeroGivPower ? (
```

MohammadPCh (changes requested):
Thanks @RamRamez 
we only need to show the toast if user has boosted projects and his GIVpower is zero

MohammadPCh (approved):
Thanks @RamRamez

Issue #2279, opened 2023-02-27
Can't sign in with new menu


RamRamez:
> Hi, I want to connect/sign in to post a project update. My wallet is connected, but the screen  with the sign in button to click shows up really briefly and than disappears. So I can not sign in. Is there some sort of bug crept in at the moment?

[Ref](https://discord.com/channels/679428761438912522/835168432520429578/1079874711548350597)

maryjaf:
Thanks @RamRamez it has been resolved

PR #2280, opened 2023-02-27
Can't sign in with new menu

RamRamez:
#2279


maryjaf:
plz check this @RamRamez @MohammadPCh 
When I am signed out and I select the "Create Project" from the menu
The window for completing the profile will be displayed
And when I click OK, loading is displayed and nothing happens

image
image
image

MohammadPCh:
We needed to fix it as soon as possible and merge it to production, so I create another pr and fix both problems.

Issue #2271, opened 2023-02-27
Frontend: Add Polygon network


MoeNick:
- adding recipient address in Create Project Page,
- switch the network in the header, @mosaeedi
- supported network error in the donation page
- unsupported network on GIVeconomy!

mosaeedi:
# Design resource

1. user can tap on the **Switch network** and open the switch network modal

### How it looks on the Desktop
[Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=20714%3A268639&t=gd5S1Nfqg0kilgTO-4)

### And Small tablets
[Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=20714%3A267784&t=gd5S1Nfqg0kilgTO-4)


### Mobile version
[Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=20714%3A268060&t=gd5S1Nfqg0kilgTO-4)


## A tiny update for the wallet/user dropdown
1. The network icon is added 
2. the font size is changed to **Body/caption-medium**

[Figma link to review the wallet/user dropdown
](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=20714%3A268489&t=gd5S1Nfqg0kilgTO-4)

<img width="261" alt="Screenshot 2023-02-28 at 2 23 06 PM" src="https://user-images.githubusercontent.com/13708322/221833276-99cd27a9-f5d4-4c42-8bef-dd22626ec34a.png">

mosaeedi:
@MoeNick design resources are added.

RamRamez:
@MoeNick 
Should I add Polygon chain to the Manage funds section in the verification steps?

MoeNick:
Yes please but only on its feature branch to test.

RamRamez:
> * adding recipient address in Create Project Page,
> * switch the network in the header, @mosaeedi
> * supported network error in the donation page
> * unsupported network on GIVeconomy!

PR #2296 
- add polygon network params
- add polygon to create project page
- add polygon chain native token (Matic) balance to donate page
- add polygon chain to donate modal to fetch polygon token prices
- add polygon logo and name to donate page wrong network modal
[...]

PR #2296, opened 2023-03-05
Add Polygon Chain

RamRamez:
#2271


MohammadPCh (approved):
it's not on our standard flow, I just merge it without review.

RamRamez:
> it's not on our standard flow, I just merge it without review.

Merging without review is against best practices and against our rules

Issue #2263, opened 2023-02-26
Adding new newsletter provider


MoeNick:
1) For the homepage, could we add this iframe? I'm not sure how it will look, but this is the one that they recommend to add:

`<iframe src="https://paragraph.xyz/@giveth/embed" width="480" height="480" style="border:1px solid #EEE; background:white;" frameborder="0" scrolling="no"></iframe>`

You can also check their docs https://docs.paragraph.xyz/docs/development

2) For the /join page, can you add the iframe or this link? https://paragraph.xyz/@giveth/subscribe

_Originally posted by @ahmadabugosh in https://github.com/Giveth/giveth-planning/issues/1094#issuecomment-1368770839_

MohammadPCh:
Hmmm It Looks like this:
![image](https://user-images.githubusercontent.com/17097576/221420479-4090db0e-83d8-49e7-a8d8-6ee25a2c7ebb.png)
We need @mosaeedi :heart:  here.

MoeNick:
As @mosaeedi suggested we decided to remove subscribe and think about adding a link to the newsletter later.

mosaeedi:
It doesn't look good, and it's not the same style as the rest of the homepage, for now, please hide the newsletter section and let's find a better solution ASAP.

ahmadabugosh:
@MoeNick @mosaeedi @MohammadPCh Can we just add a button for people that opens in a new tab for them to subscribe? https://paragraph.xyz/@giveth/subscribe

MoeNick:
@mosaeedi is working on it, maybe with a modal.

mosaeedi:
# Design resource
## Newsletter section

![Screenshot 2023-02-28 at 3.15.08 PM.png](https://images.zenhubusercontent.com/61dd85b9a1419a0fc83f15d1/031a1612-361f-4dbc-af75-d256247300c0)

### (>1280 & <1439)
[Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=20714%3A283507&t=gd5S1Nfqg0kilgTO-4)

### (>1024  & <1279)
[Figma link](https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/Giveth.io-%26-GIVeconomy?node-id=20714%3A283555&t=gd5S1Nfqg0kilgTO-4)

### (>768  & <1024)
[...]

PR #2298, opened 2023-03-06
adding new newsletter provider to home and join page

RamRamez:
#2263


src/components/views/homepage/introBlock/index.tsx
```diff
@@ -186,6 +185,20 @@ const IntroContainer = styled.div`
 	padding: 0 8px;
 `;
 
+const ContainerStyled = styled(Container)`
+	position: relative;
+	${mediaQueries.tablet} {
+		padding: 0;
+	}
+`;
```
MohammadPCh, line 193: Thanks @RamRamez 
It's our standard grid, we don't need to remove its padding.
```suggestion
const ContainerStyled = styled(Container)`
	position: relative;
`;
```

MohammadPCh (changes requested):

src/components/views/homepage/introBlock/index.tsx
RamRamez: I fixed tablet screen size

image

it's according to figma
https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/GIVeconomy?node-id=17955%3A234602&t=PCNLkDk91cgonIv8-0

MohammadPCh (approved):
Thanks @RamRamez

Issue #1860, opened 2022-11-22
"Sign out" Option does work.


laurenluz:
When I try to "Sign out" via our drop down menu (from the top where my name is)... the "sign out" actually doesn't disconnect my wallet or sign me out. It just makes the "sign out" option disappear.

@mateodaza @RamRamez \

![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/72115a0e-22d0-4657-b58d-b47bb23b558a)

RamRamez:
I tested with MetaMask and Brave and it was ok. @laurenluz which wallet are you using?

RamRamez:
shields are up for me and it's ok.
@laurenluz what do you mean by `"sign out" actually doesn't disconnect my wallet or sign me out`? What is happening when you sign out?

laurenluz:
@RamRamez Well, usually, after a 1st login, in this drop down menu I see a "sign out" option at the bottom. After I click it.. the page kind of refreshes & the sign out button disappears... but I'm still signed in:

![image](https://user-images.githubusercontent.com/75490971/204356209-3e924b7f-6b6b-4358-9ef4-b1a224ed2b15.png)

MoeNick:
THis is the old ux problem we have for sign-in with eth,
currently: 
1- sign-out will not disconnect you from DApp, a wallet is still connected and we know who logged in, but it expires the token you signed in before with your wallet. so you get sign-out, you don't have a token and if you want to like a project, it asks you to sign again. Apparently in this situation, you can't see the sign-out choice. because you signed out already. 

2- Actually connecting the wallet wont mean logged in , we ask user to log in again with a msg to the Metamask or a wallet, so basically it's a 2 step connecting and sign the logging msg with a user wallet,. since the beginning we have this argument with design team that this flow won't mean the user signed in, we compromised it. 

However. we are solving this problem in MLA with @mosaeedi

laurenluz:
Yeah I talked about this with @RamRamez in DM. The average user doesn't understand anything about how sign out deletes the token... in a web2 application sign out means your account isn't connected anymore. 

I think it would be better if sign out deleted the token & disconnected the wallet simultaneously.

MoeNick:
is it possible to disconnect from DApp @RamRamez ?

RamRamez:
> is it possible to disconnect from DApp @RamRamez ?

Yes, It's possible.
I've prepared a demo:
[...]

PR #2300, opened 2023-03-06
Automate login flow - connect wallet then sign in then do the desired action

RamRamez:
[#1860](https://github.com/Giveth/giveth-dapps-v2/issues/1860#issuecomment-1455988104)


MohammadPCh (approved):
Thanks @RamRamez

Issue #2306, opened 2023-03-08
Homepage Performance Improvements


alireza-sharifpour:
![Screen Shot 2023-03-08 at 12 31 05 PM](https://user-images.githubusercontent.com/48887817/223669161-b332e275-1fc0-4afb-9afa-876b43a3d5ec.png)

This is the report we get from GTmetrix, i think we can improve it with some tiny changes. we can work on improvement in background images and particles and other stuff that GTmetrix suggests.

https://gtmetrix.com/reports/giveth.io/axhwhWnL/

MohammadPCh:
@alireza-sharifpour Good point. Thanks for reporting this.

maryjaf:
How can I test? QA test is needed for this issue?  @alireza-sharifpour

PR #2311, opened 2023-03-09
Change intro block bg

alireza-sharifpour:
- #2306


MohammadPCh (approved):
Thanks @alireza-sharifpour

PR #2314, opened 2023-03-09
fix latest updates images

MohammadPCh:
- Giveth/impact-graph#856


maryjaf:
In the list of latest updates, only one update from each project should be displayed? @MoeNick  
For example, if we have an update in the list for a project, if the second one is registered, the previous one will be displayed.

Issue #2045, opened 2023-01-18
Giveth TRACE deprecation - we should adjust copy on Giveth.io in FAQ & history


laurenluz:
The copy on our dApp should be adjusted to refect the fact that Giveth TRACE has been deprecated. We can still mention it as part of our history, but it should be past-tense (like Giveth TRACE was built for traceability, and since has been deprecated because of the merge & rinkeby depreaction...)

This issue first requires someone from comms to go through the dApp copy & find what needs to be changed, and then make suggested changes (that are then reviewed).
Main things to look at are:
- FAQ
- About
- History

@c0ric0ri - this might be a good issue for you, wdyt?

Reference #1877

laurenluz:
@danelsuga - can you take this on?

danelsuga:
Is the Trace button supposed to still be there? Nothing loads on the page (errors all around), but it still exists. Feels confusing both on the level of the presence of the button on the homepage and on the Trace page itself.


FAQ's  https://giveth.io/faq

What is a Traceable project?
**CHANGE TO:** 
What was a Giveth “Traceable” project?

Giveth TRACE was the original fundraising application of Giveth that was sunsetted in Q3 2022 at the time of the ETH Merge and Rinkeby network deprecation. Its foundation was traceability and hence offered Verified projects on [Giveth.io](https://giveth.io/) to choose to also become a [Campaign](https://docs.giveth.io/dapps/entitiesAndRoles/#campaigns) on [Giveth TRACE](https://trace.giveth.io/), thus becoming a Traceable Project. It enabled project creators to manage their donations transparently using [Traces](https://docs.giveth.io/dapps/entitiesAndRoles/#traces). One
feature of Traces was specifying how a project used their donations to achieve the goals of the overarching Campaigns. Upgrading Giveth.io projects to Campaigns enabled project owners to specify parts of their project requiring funding as different types of [Traces](https://docs.giveth.io/dapps/entitiesAndRoles/#traces). Donors benefited from being able to choose to fund either specific Traces or the overarching Campaign, and were thereby able to trace the flow of their donations. Traceable projects formerly appeared on both [Giveth TRACE](https://trace.giveth.io/) (as a Campaign) and [Giveth.io](https://giveth.io/), creating double exposure. To learn more about the history Giveth TRACE and  how projects were made traceable, visit our [documentation](https://docs.giveth.io/dapps/makeTraceableProject).

danelsuga:
[...]

PR #2329, opened 2023-03-15
Update trace copies

alireza-sharifpour:
- #2045


MohammadPCh (approved):
Thanks @alireza-sharifpour

PR #2360, opened 2023-03-18
remove links to access to minting page

divine-comedian:
prepare the page ahead of early minting to allow users to check their addresses but not mint


mateodaza (approved):
lgtm!

Issue #2363, opened 2023-03-20
Move Grid to Design system


alireza-sharifpour:
We need to move our Grid and its related components (Container, Col, Row, ...) to our design system.

FYI: @MohammadPCh, @RamRamez, @mateodaza

PR #2364, opened 2023-03-20
updated all Grids to be used from Design system and remove project's …

alireza-sharifpour:
- #2363


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #2371, opened 2023-03-20
Resize Notification Page header 


MoeNick:
The padding in design is set to 70, the padding right now is 170. 

![image.png](https://images.zenhubusercontent.com/604f242c2819d5da6426ad03/20f4b366-5342-475d-8d11-0a611bcdb2ba)

![image.png](https://images.zenhubusercontent.com/604f242c2819d5da6426ad03/6c3ded8c-7c59-43f7-b70a-3765ff682f1c)

PR #2373, opened 2023-03-20
fixed margins

alireza-sharifpour:
- #2371


MohammadPCh (approved):
Thanks @alireza-sharifpour

PR #2387, opened 2023-03-21
add walletAddress in donation table for users with no name

RamRamez:
[Ref.](https://github.com/Giveth/impact-graph/issues/872)


mateodaza (approved):
looking good, thanks man

PR #2403, opened 2023-03-23
remove checkeligibility after 24th March 2023, 11am CST

alireza-sharifpour:


MohammadPCh:
Thanks, @alireza-sharifpour 
we can have an easier solution, and remove it at a specific time, instead of checking the time each time the user opens that page. because we should check the time by the server and don't rely on the user's system time.

Issue #2416, opened 2023-03-28
Fix Copy & rearrange text in projects tab


laurenluz:
This is so messy, let's make some small changes

![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/b538eec8-5b02-41cc-be4f-452a4c453bcc)

- [x] "All projects" should be 1st
- [x] "Recently Updated" needs a space
- [x] Second title should just say "CATEGORY" (no "by")

@MoeNick

laurenluz:
this is @mosaeedi approved!

Can @RamRamez or @alireza-sharifpour  take it?

alireza-sharifpour:
> this is @mosaeedi approved!
> 
> Can @RamRamez or @alireza-sharifpour take it?

I'll take it.

PR #2425, opened 2023-03-28
fixed_typos

alireza-sharifpour:
- #2416


MohammadPCh (approved):
Thanks @alireza-sharifpour

PR #2483, opened 2023-04-09
Update catalan

sembrestels:
I did some fixes after reviewing manually some of the language keys.


mateodaza (approved):
you rock, thanks a lot @sembrestels

PR #2516, opened 2023-04-13
fixed rerender and width issues

alireza-sharifpour:


MohammadPCh (approved):
Thanks @alireza-sharifpour

PR #2519, opened 2023-04-14
image sizes and text improvements

divine-comedian:


divine-comedian:
I made some improvements to the image sizes and the text descriptions we use. 

I also noticed one bug but it was out of my skill level to fix it - could either of you check it @alireza-sharifpour @MohammadPCh ?

On the single project page with smaller screens the tooltip text is cut off from hovering the pfp badge
image

MohammadPCh (approved):
Thanks @divine-comedian, The hover animation is really eye-catching. Great Job.

MohammadPCh:
@divine-comedian Thanks, Yes, unfortunately, our tooltip does not support this feature. maybe we can fix it in the future. but for now we can easily change the tooltip position to the right :)

PR #2536, opened 2023-04-19
Add link to account page and fix max nft amount

alireza-sharifpour:


RamRamez (approved):
thanks @alireza-sharifpour

PR #2566, opened 2023-04-23
Praise support in Stream

aminlatifi:
image

@divine-comedian An appropriate icon must be designed and then added by the frontend team.


MohammadPCh:
Result:
image

MohammadPCh (commented):
Thanks @aminlatifi

Issue #2377, opened 2023-03-20
Show Giveth PFP users on the dapp


divine-comedian:
As part of the second part of PFP functionality #2319. Once a user sets their profile picture they should have a small "pfp badge" that show up in the following places:

- next to project owner in project card
- next to project owner name on single project page 
- next to donor name in project donations tab
- next to booster name in project givpower tab

![image](https://user-images.githubusercontent.com/67759413/226376831-a6e59c74-8d02-46e7-92d4-37e975158f87.png)

You can check it out here:
https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/GIVeconomy?node-id=14997-268092&t=DQNhwzQWgTUuJMyu-0

When user hovers over one of these badges we should show some text as a hover tooltip. when the user clicks on one of these badges we open a new tab taking them to the NFT page on opensea for that specific NFT

divine-comedian:
some updates from the call today:

adding a hover effect on pfp badges located on the project card and next to project owner name on the single project page
when a user hovers over the badge we should increase the glow effect and show the same hover tooltip we show on hover in the donation table.

![image](https://user-images.githubusercontent.com/67759413/229859858-bbb7cfa8-a1d5-4fbc-a5e3-021113550604.png)


- [ ] Add hover effects to project card for pfp badges
- [ ] add hover effects to project owner pfp badges on single project page

divine-comedian:
testing on https://givpfp.vercel.app/

We should fix the current flow for when a user hovers over a PFP image:
![image](https://user-images.githubusercontent.com/67759413/231219247-a96d0261-4063-4983-8883-3ecd3eb6703e.png)



We're going to change the flow a little bit here.
While the collection has not yet sold out we should redirect users to the nft page. - https://giveth.io/nft

Currently when I click on the tooltip it takes me to the user's profile page..

When the collection sells out we should change this text and also the link to take them to the Rarible page for the specific NFT.
example - https://rarible.com/token/0x78fde77737d5b9ab32fc718c9535c7f1b8ce84db:98?tab=overview

divine-comedian:
Hey @MohammadPCh 

As per the Figma, on the user's profile page we should add the red glow and drop shadow around the user's NFT PFP

### Current
![image](https://user-images.githubusercontent.com/67759413/232113062-24ef66b4-1aa6-45d1-b080-dc70cddc8516.png)

### Figma
[...]

PR #2572, opened 2023-04-25
moved tooltip to right

alireza-sharifpour:
- #2377


MohammadPCh (approved):
Thanks @alireza-sharifpour

PR #2574, opened 2023-04-25
test pfp on mainnet

alireza-sharifpour:


MohammadPCh:
I think we don't need it anymore. thanks @alireza-sharifpour

PR #2577, opened 2023-04-25
last minute ux improvement

divine-comedian:
This was a last minute UX suggestion from @laurenluz 

If the user holds a Givers NFT in their wallet, is looking at their profile page and has not set a PFP as their profile picture we suggest them to do so and provide a direct link to the setProfilePic Modal. 

If the user does not hold a Givers PFP we show nothing

If the user is viewing any profile, including their own that has a pfp set already as their profile pic then we show a link to view it on Rarible.


alireza-sharifpour (approved):
Thanks @divine-comedian

Issue #2565, opened 2023-04-23
Public view| Similar projects


MoeNick:
![Screenshot 2023-05-01 at 4.38.38 PM.png](https://images.zenhubusercontent.com/61dd85b9a1419a0fc83f15d1/de8904d9-e03e-465c-9cbd-497717b6f920)

maryjaf:
I checked on https://giveth-dapps-v2-git-2565newprojectviewsimilarprojects-givethio.vercel.app/
but I didn't see any changes in similar project view @RamRamez Am I checking the right place?
<img width="1424" alt="image" src="https://user-images.githubusercontent.com/111529185/236620324-27e74323-47bc-44fd-b83b-2ba8a701d76c.png">

RamRamez:
@maryjaf Some colors, padding, and margins have changed a button is added.
![image](https://user-images.githubusercontent.com/73441857/236623739-48201c99-db09-46c3-bc74-b08a9608a202.png)

You can compare with Figma design
https://www.figma.com/file/C3HykACu0eNImhAqu0our8/Giveth-projects?type=design&node-id=131-60706&t=6vxWavpq6Lu2OaJu-0

maryjaf:
> @maryjaf Some colors, paddings, and margins have changed and a button is added. ![image](https://user-images.githubusercontent.com/73441857/236623739-48201c99-db09-46c3-bc74-b08a9608a202.png)
> 
> You can compare with Figma design https://www.figma.com/file/C3HykACu0eNImhAqu0our8/Giveth-projects?type=design&node-id=131-60706&t=6vxWavpq6Lu2OaJu-0

Thanks @RamRamez I see differences between figma and  feature branch, plz check
and I have one guestion 
![image](https://user-images.githubusercontent.com/111529185/236660642-1e6520df-1ab1-48e0-b258-1ff092cf4330.png)![image](https://user-images.githubusercontent.com/111529185/236660651-972d44a3-5150-4cbf-ae04-caf65b694a70.png)

RamRamez:
> > @maryjaf Some colors, paddings, and margins have changed and a button is added. ![image](https://user-images.githubusercontent.com/73441857/236623739-48201c99-db09-46c3-bc74-b08a9608a202.png)
> > You can compare with Figma design https://www.figma.com/file/C3HykACu0eNImhAqu0our8/Giveth-projects?type=design&node-id=131-60706&t=6vxWavpq6Lu2OaJu-0
> 
[...]

PR #2619, opened 2023-05-07
Fix author color

RamRamez:
https://github.com/Giveth/giveth-dapps-v2/issues/2565#issuecomment-1537364147


alireza-sharifpour (approved):
Thanks @RamRamez

Issue #2647, opened 2023-05-12
Sometimes the "All notification" doesn't shown in notification menu


maryjaf:
**Describe the bug**
When the text of 4 notifications displayed in the menu is more than a certain limit, "All notification" is not displayed and the user couldn't tap on "all notification" to see all o them
@MohammadPCh 

**Screenshots**
<img width="1433" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/6f42cefe-35c9-441a-9ad5-a605030f1185">

MohammadPCh:
@alireza-sharifpour can you check it, please?

alireza-sharifpour:
> @alireza-sharifpour can you check it, please?

Sure, i'm working on it

PR #2666, opened 2023-05-16
Fix notification link issue

alireza-sharifpour:
#2647


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #2313, opened 2023-03-09
Too much information in the first steps to verify a project


mosaeedi:
We show too much information before starting the verification process!

By tapping on the verify project, 
user will present with this modal

![Screenshot 2023-03-09 at 12.46.41 PM.png](https://images.zenhubusercontent.com/61dd85b9a1419a0fc83f15d1/d6aa8fdf-72fe-4196-bd80-6286c760fce7)

Then, when they read all those texts, we present the verification form with too much text to read again!

![Screenshot 2023-03-09 at 12.53.51 PM.png](https://images.zenhubusercontent.com/61dd85b9a1419a0fc83f15d1/15437fb1-efc4-4eb9-903b-c6fe1be5d056)

---

It's not a good user flow for someone who starts the verification process.

WhyldWanderer:
We can remove the highlighted text
![image](https://user-images.githubusercontent.com/74267818/224212801-79ac3f80-0600-4d32-a48e-86b87d21f558.png)

WhyldWanderer:
Also here we can remove the highlighted text and the text circled in red should link to https://docs.giveth.io/dapps/projectVerification/

![image](https://user-images.githubusercontent.com/74267818/224213020-a904a1e6-c0d3-46fa-bdbb-b659229a43bb.png)

PR #2672, opened 2023-05-17
Verification text edits

alireza-sharifpour:
#2313


RamRamez (approved):
Thanks @alireza-sharifpour

PR #2700, opened 2023-05-23
Changing thegraph from "hosted service" to "graph studio"

geleeroyale:
~~Not sure if this is the correct change. We technically have no staging environment deployed on studio. Reusing the production gnosischain one in `development.ts`.
`production.ts`should be good with the changes made.~~

I added a staging deployment.

### DIFF
--- a/src/config/development.ts
+++ b/src/config/development.ts
@@ -47,7 +47,7 @@ const config: EnvConfig = {
 		},
 		blockExplorerName: ['Etherscan'],
 		subgraphAddress:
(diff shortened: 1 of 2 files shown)


geleeroyale:
In order to be able to have a structured discussion and transparency with this migration I prepared a document. Please request access from me if you need it:

https://www.notion.so/giveth/Subgraph-Studio-8a497caa2d6a43cb871cf8919f8e8bbd?pvs=4#8a9d49291eb244cb8f457bbeae225914

Issue #2651, opened 2023-05-12
QF - Add Quadratic Funding to Projects drop-down navigation


laurenluz:
To get to the Quadratic Funding project page, the user can navigate through the drop down menu from "Projects". This will go to a new page showing only the projects & details that are part of QF.

I'd love if the URL for that new page could be https://giveth.io/qf

![image](https://github.com/Giveth/giveth-dapps-v2/assets/75490971/01c44d50-2ab7-4de8-8659-817a300d57fe)

Link to [figma](https://www.figma.com/file/C3HykACu0eNImhAqu0our8/Giveth-projects?type=design&node-id=922-61393&t=v9RpJgXeL6iRVDHB-4)

@MoeNick @MohammadPCh - can you assign the appropriate person?

cc @markoprljic @mosaeedi

MoeNick:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/All-Devs-61644589b92edf0016f6999c/issues/Giveth/giveth-dapps-v2/2651?planning-poker) with Zenhub @alireza-sharifpour @mateodaza @MohammadPCh @RamRamez

maryjaf:
As a suggestion 
Should an item related to QF be added on this search page? @laurenluz 
![image.png](https://images.zenhubusercontent.com/62fe7a8b57c1a1770575ea40/58f7c60c-5500-43a3-b07e-51be3501f53a)

maryjaf:
Plz check this scenario @RamRamez 
When I tap on "Quadratic funding" and this link is shown https://givqf.vercel.app/qf after that by taping on "All projects" in drop down menu nothing happened @RamRamez

RamRamez:
> Plz check this scenario @RamRamez When I tap on "Quadratic funding" and this link is shown https://givqf.vercel.app/qf after that by taping on "All projects" in drop down menu nothing happened @RamRamez

Thanks @maryjaf I fixed it.

laurenluz:
> As a suggestion
> Should an item related to QF be added on this search page? @laurenluz

I think it's a good idea as a quick link - but it should only be there when there is an active QF round @maryjaf - can we do that?

maryjaf:
> > Plz check this scenario @RamRamez When I tap on "Quadratic funding" and this link is shown https://givqf.vercel.app/qf after that by taping on "All projects" in drop down menu nothing happened @RamRamez
> 
> Thanks @maryjaf I fixed it.

[...]

PR #2707, opened 2023-05-28
Fix projects route

RamRamez:
[2651](https://github.com/Giveth/giveth-dapps-v2/issues/2651#issuecomment-1565899666)


MohammadPCh (approved):
Thanks @RamRamez

Issue #2653, opened 2023-05-12
QF - Project Single page for QF projects


laurenluz:
For projects that are actively part of a QF round, they will have a modified project single page. It will look like this:

![image](https://github.com/Giveth/giveth-dapps-v2/assets/75490971/c12a8b4b-8f29-475e-8a7d-1ec3bf9bd71d)

Notes:
- The key difference is the donate card w/ the estimate matching amounts & numbers
- The numbers for the estimate matching on donations should come from the work in: https://github.com/Giveth/impact-graph/issues/980 (forked from gitcoin)
- The gitcoin passport banner will change in styling and will have different variations on text
- Donations page is not finished yet, we can only work on the main tab

Link to [figma](https://www.figma.com/file/C3HykACu0eNImhAqu0our8/Giveth-projects?type=design&node-id=1331-59634&t=v9RpJgXeL6iRVDHB-4)

Can you assign the appropriate person? @MoeNick @MohammadPCh 

cc @markoprljic @mosaeedi

laurenluz:
Note: gitcoin passport banner is in #2654

MoeNick:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/All-Devs-61644589b92edf0016f6999c/issues/Giveth/giveth-dapps-v2/2653?planning-poker) with Zenhub @alireza-sharifpour @mateodaza @MohammadPCh @RamRamez

MoeNick:
According to @MohammadPCh it's blocked by responsive and @mosaeedi is working on this today.

MoeNick:
@alireza-sharifpour is taking care of responsiveness of this page

alireza-sharifpour:
We don't have the backend side. @MoeNick

MoeNick:
Just to confirm @laurenluz 
![image](https://github.com/Giveth/giveth-dapps-v2/assets/80677512/c06c7645-1b51-41fd-9cf6-1aeb88c8b3f7)
These numbers are for the current round.

laurenluz:
yes exactly @MoeNick

laurenluz:
Hey @alireza-sharifpour - the designs have been updated based on our discussions. Can you revise?

[...]

PR #2746, opened 2023-06-08
Update action card

alireza-sharifpour:
#2653


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #2692, opened 2023-05-22
QF - Donation Page


laurenluz:

laurenluz:
Design is ready for the donate page! [link to figma](https://www.figma.com/file/C3HykACu0eNImhAqu0our8/Giveth-projects?type=design&node-id=2259-93618&t=H1s3nHekMcuH4OGd-4)

@MohammadPCh - can you assign the right person?

maryjaf:
When I enter amount of GIV, estimated matching is shown 0 ?why? @RamRamez 
![image.png](https://images.zenhubusercontent.com/62fe7a8b57c1a1770575ea40/db913d64-24ce-4324-85b7-74c6f89cd199)

mosaeedi:
## Responsive guides

Donation (>1280 & <1439)
[Figma link](https://www.figma.com/file/C3HykACu0eNImhAqu0our8/Giveth-projects?type=design&node-id=3344-116619&mode=design&t=ZFYDlXVbvJeWRFXN-4)

Donation (>1024  & <1279)
[Figma link](https://www.figma.com/file/C3HykACu0eNImhAqu0our8/Giveth-projects?type=design&node-id=3344-116711&mode=design&t=ZFYDlXVbvJeWRFXN-4)

Donation (>768  & <1024)
[Figma link](https://www.figma.com/file/C3HykACu0eNImhAqu0our8/Giveth-projects?type=design&node-id=3344-116881&mode=design&t=ZFYDlXVbvJeWRFXN-4)

Donation (Mobile)
[Figma link](https://www.figma.com/file/C3HykACu0eNImhAqu0our8/Giveth-projects?type=design&node-id=3344-118355&mode=design&t=ZFYDlXVbvJeWRFXN-4)

RamRamez:
> @maryjaf I tested for GIV and WETH tokens and the estimated matching changes as expected. But it doesn't change when I enter different amounts of xDai. @alireza-sharifpour can you check it please?

I fixed it and sent the PR. It's waiting for review.
@maryjaf you can test it after the PR is merged.

maryjaf:
> > @maryjaf I tested for GIV and WETH tokens and the estimated matching changes as expected. But it doesn't change when I enter different amounts of xDai. @alireza-sharifpour can you check it please?
> 
> I fixed it and sent the PR. It's waiting for review. @maryjaf you can test it after the PR is merged.

ُThanks, Please let me know when it is ready to test

[...]

PR #2765, opened 2023-06-14
Add estimatedmatching donation

alireza-sharifpour:
#2692


RamRamez (approved):
Thanks @alireza-sharifpour

PR #2785, opened 2023-06-22
Use subgraph studio endpoints on production frontend

geleeroyale:
These should not be hard coded in here as we are now exposing our API key. However - the key is limited from thegraph dashboard

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "givethdapp",
-	"version": "2.15.9",
+	"version": "2.15.10",
 	"private": true,
 	"scripts": {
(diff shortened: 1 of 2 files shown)


MohammadPCh:
@geleeroyale Thanks, it's actually difficult to hide this from the user because we need to use it to send requests. Since we must have it in the client, the user will be able to see it.

MohammadPCh (approved):
Thanks @geleeroyale

geleeroyale:
I think we need to roll this one back. Need to work together with devs as switching from hosted to studio subgraphs necessitates some changes it seems

Issue #2693, opened 2023-05-22
QF - Donation Success Page


laurenluz:

maryjaf:
![image.png](https://images.zenhubusercontent.com/62fe7a8b57c1a1770575ea40/23c97136-e5bb-42ab-8de2-1a2d8b204a43)

I don't know this time is related about which fields in backend , It seems wrong @RamRamez

RamRamez:
> ![image.png](https://camo.githubusercontent.com/6e8f4eb022d2c66d0a89c6d81c3f2b52146b619edbe832e89af5e69b95e46c8f/68747470733a2f2f696d616765732e7a656e68756275736572636f6e74656e742e636f6d2f3632666537613862353763316131373730353735656134302f32336339373133362d653562622d343261622d386465322d316132643862323034613433)
> 
> I don't know this time is related about which fields in backend , It seems wrong @RamRamez

I fixed it and sent the PR. It's waiting for review.

maryjaf:
When I tap on "Go to passport" I'm redirected to :https://givqf.vercel.app/ @RamRamez

maryjaf:
Plz check this scenario, is this expected? @RamRamez 
I donate with a wallet which it's score is lower than minimum score and after donation I see this box correctly
![image.png](https://images.zenhubusercontent.com/62fe7a8b57c1a1770575ea40/6b329ca9-bc70-44b8-a6e2-08f87f985488)
when I change my wallet from MetaMask and connect a wallet with higher score, the box is changed although I didn't donate with this wallet 
![image.png](https://images.zenhubusercontent.com/62fe7a8b57c1a1770575ea40/8d53d78e-17db-4425-a94d-0a608768a15b)
and some times the end date isn't shown correctly
![image.png](https://images.zenhubusercontent.com/62fe7a8b57c1a1770575ea40/8caad848-ed48-4406-9edf-87c6cedd0527)

RamRamez:
> When I tap on "Go to passport" I'm redirected to :https://givqf.vercel.app/ @RamRamez

I fixed it and my PR is waiting for review.

RamRamez:
[...]

PR #2786, opened 2023-06-22
Fix end date

RamRamez:
https://github.com/Giveth/giveth-dapps-v2/issues/2693#issuecomment-1600906413


MohammadPCh (approved):
@RamRamez Thanks

PR #2791, opened 2023-06-25
fix passport link

RamRamez:
https://github.com/Giveth/giveth-dapps-v2/issues/2693#issuecomment-1605885623


MohammadPCh (approved):
Thanks @RamRamez

PR #2794, opened 2023-06-26
fix go to passport link to internal link

RamRamez:
https://github.com/Giveth/giveth-dapps-v2/issues/2693#issuecomment-1608124977


MohammadPCh (approved):
Thanks @RamRamez

Issue #2635, opened 2023-05-09
404 error


brichis:
GM @MoeNick ! I think I found a bug 

**Describe the bug**
When I'm on my profile, and I click My GIVpower, My Projects or My Donations, it shows me an error but I can access normaly to my project if I enter directly from its link.

**To Reproduce**
1. Go to https://giveth.io/account?data=all&view=projects
2. Click on My GIVpower, My Projects or My Donations
3. See error

**Screenshots**

https://github.com/Giveth/giveth-dapps-v2/assets/106570342/d97de48d-7ef8-400c-99cb-a0447a5de9de

**Desktop (please complete the following information):**
 - OS: iOs
 - Browser Brave
 - Version 1.51.110

**Additional context**

MoeNick:
Thanks for reporting
@maryjaf  can you check all scenarios for thiss issue

Vyvy-vi:
@MoeNick This is still broken. Can I work on this?

MoeNick:
we could not reproduce it, if you could, yes please take it.

MohammadPCh:
@Vyvy-vi we should change our logic in : ProfileContributes.tsx
```
<Link href={pathname + profileTabs.overview}>
	<ProfileTab active={tab === EProfile.OVERVIEW}>
		{formatMessage({ id: 'label.overview' })}
	</ProfileTab>
</Link>
```
we have to add query params in another way :)
can you take a look?

divine-comedian:
@MohammadPCh is this still relevant?

maryjaf:
It is an old problem, now there is no bug in this page.

PR #2814, opened 2023-07-05
fix: url in removeQueryParam

Vyvy-vi:
related to #2635 

This doesn't solve the issue completely but it fixes a minor bug with the formatted string in `removeQueryParam`


MohammadPCh (approved):
Thanks @Vyvy-vi , Nice catch

PR #2822, opened 2023-07-07
Upgrading web3 react from v6 to v8

mateodaza:


MohammadPCh:
I think we can close it :)

PR #2890, opened 2023-07-12
delete redundant statement from givpower rank

### LINKED ISSUE #2864: Project single page - GIVpower tab - delete redundant statment on GIVpower rank card
We should remove the part about "this rank will update at the start of the next GIVbacks round"... because we are showing the date and it is redundant.

@MohammadPCh

Vyvy-vi:
fixes #2864


MohammadPCh (approved):
Thanks @Vyvy-vi

PR #2894, opened 2023-07-13
Link to donor profile on project donations tab

### LINKED ISSUE #612: Donors' names on Donation tab of a project
Shouldn't the names of the donor be clickable if they have an existing account/project in the site? I'm not sure if this is important but I'm just mentioning it. 

image

This is the page where I got the screenshot above:
https://giveth-dapps-v2.vercel.app/project/ram-second-torus 

@MoeNick

Vyvy-vi:
fixes #612 

We should probably update the design for this a bit to make it more obvious that there ae two clickable in that components -> link to the block explorer for that transaction, and the user profile, instead of highlighting the whole component on hover...


MohammadPCh (approved):
Thanks @Vyvy-vi

Issue #2896, opened 2023-07-13
QF - Estimated matching calculations - can we round to the nearest whole number please? :)


laurenluz:
It looks very strange that it's estimated like pennys of matching.. but the numbers are all totally made up... can we round to the nearest whole number please?

So we'd see...

$367
$83
$154

etc.

... for the estimated matching numbers? @MohammadPCh @RamRamez @alireza-sharifpour - who can take it? :)

![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/91686bb0-ab2c-461c-bfde-2899848c38aa)

MohammadPCh:
Yes, let's do it :)

MohammadPCh:
@laurenluz please test it again, thanks.

laurenluz:
it's better - but we're still missing the "$" on $0 when there is no matching allocated

![image](https://github.com/Giveth/giveth-dapps-v2/assets/75490971/c97b1224-4414-4958-8b6b-3dc00e74b396)

maryjaf:
> 1. For QF projects that have no donations yet, we should show estimated matching $0

This Item is not implemented @MohammadPCh 
<img width="480" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/03eb0090-e85d-46e6-8e82-e57d6506a09b">

PR #2901, opened 2023-07-13
Fix calculate total estimated matching

alireza-sharifpour:
#2896


laurenluz:
@alireza-sharifpour - can we include the "$" even on "0" value estimated matching?

Issue #2912, opened 2023-07-14
Successful donation w/ completed passport shows CTA to go to passport..


divine-comedian:
Staging
Brave Browser

Using an account with an already approved passport.
On making a successful donation to a QF eligible project the main CTA is for me to go to the passport page. If I already have a passport connected with a good score then why should I need to go back to this page? 

This should take me back to the list of QF eligible projects to make another donation, not to the passport page 

![image](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/1a6c8e37-3c2c-4afd-8060-57df60fabd5c)

laurenluz:
yeah good call... with this condition (when the user makes a donation and they have sufficient passport score), I think we can make the button "Go to Projects" and have it link to giveth.io/qf

we have to be careful that this change doesn't affect the "go to passport" button for the other conditions (no passport, passport score not high enough)

@MohammadPCh @divine-comedian

laurenluz:
cc @maryjaf @MoeNick

MohammadPCh:
@RamRamez can you take a look?

MohammadPCh:
@maryjaf It's ready to test.

maryjaf:
> @maryjaf It's ready to test.

<img width="1303" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/19391e78-980b-47bb-9884-e0349e6a9909">
I'v checked and still "Go passport "is shown, plz check @MohammadPCh

RamRamez:
It takes about 2 or 3 seconds for the toast to update, so I added a condition, we only show the toast when it's ready.

MohammadPCh:
Merged and ready to test in 5 mins. @maryjaf

laurenluz:
@maryjaf @divine-comedian - can you test this on staging?

divine-comedian:
looks great! testing on staging.
![image](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/ae351db6-44ba-4d4b-92da-c5795d3af8a6)

PR #2921, opened 2023-07-16
"Go to Projects" instead of passport for eligible QF donors

RamRamez:
#2912


MohammadPCh (approved):
LGTM, Thanks @RamRamez

Issue #2854, opened 2023-07-11
Add label or clear marker for QF projects to make them look different


laurenluz:
Many people in the testing parties said they couldn't tell the different between QF projects and not QF projects.

@mosaeedi @markoprljic - can we add something? like a label or badge? it's ok if it's redunant... I think we need to be clear.

we should perhaps put the same kind of badge on the project single page. because it will be disappointing for people to donate to not-QF projects without realizing it.. we need to make it clear

mosaeedi:
We can use tags,

QF Tag
Content: QF Alpha Round
Font color: Neutral/Gray/100
Background: Secondary/Cyan/600

### QF tag on single project page
<img width="1028" alt="Screenshot 2023-07-11 at 6 17 31 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/13708322/65b071d8-bbd7-4ac0-8751-8e58906d7f92">


### QF tag on Project card
<img width="635" alt="Screenshot 2023-07-11 at 6 22 45 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/13708322/82744964-243c-425d-9386-d5895f381944">

laurenluz:
@mosaeedi a few requests:
- can we remove the "GIVbacks" label? it's redundant w/ verified
- can we make "QF Alpha Round" bigger/more bold
- can we change copy "Quadratic Funding" to "Eligible for Matching"

mosaeedi:
<img width="714" alt="Screenshot 2023-07-12 at 12 09 49 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/13708322/308499d9-4dba-495d-8c26-32c9ad3b1c3c">


<img width="490" alt="Screenshot 2023-07-12 at 12 10 20 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/13708322/f6b110d7-6ede-41f1-aa62-4a57e27da261">


<img width="962" alt="Screenshot 2023-07-12 at 12 12 30 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/13708322/8e362d8c-e5cd-4273-905f-bdb163a32892">

MohammadPCh:
@laurenluz Can we start implementation? :eyes:

laurenluz:
yep, let's go! @MohammadPCh

laurenluz:
FYI @MohammadPCh - the label on the project card should be = the name set by the admin for the round in admin bro

[...]

PR #2926, opened 2023-07-17
fixed badges texts

alireza-sharifpour:
#2854


MohammadPCh (approved):
Thanks @alireza-sharifpour

PR #2928, opened 2023-07-17
added symbol for zero amount

alireza-sharifpour:
#2896


MohammadPCh (approved):
Thanks @alireza-sharifpour

PR #2960, opened 2023-07-18
Update translations

MohammadPCh:
Hey @Zeptimus, could you please take a look at these translations?
we have 2 non-translated file and 2 translated file:
- nt_es.json -> t_es.json
- nt_ca.json -> t_ca.json

nt: not translated
t: translated 
Thanks


alireza-sharifpour (approved):
Thanks @MohammadPCh

Issue #2613, opened 2023-05-04
User feedback: Confusion about multiple addresses


WhyldWanderer:
I have received feedback from users indicating that they are not understanding the process for adding wallet addresses in the Project Creation flow. Now that we have Celo network, this is a bit dangerous because most multi-sig wallets cannot be deployed with the same address on both Mainnet & Celo the way they can with Mainnet & Gnosis. 

Im not really sure how we can make this more clear for the user.. it seems clear to me.
![image](https://user-images.githubusercontent.com/74267818/236300363-39288d12-9c49-4c84-8228-e412b1e1b449.png)

I wonder if we can make this box unchecked by default?
That way they would visually see that each chain can have a different address and if they want to use the same address for all chains they can check the box?

![image](https://user-images.githubusercontent.com/74267818/236301195-d709ac1c-c588-438a-b7c4-f5bac2baa036.png)

It has happened to multiple users since adding Celo. It is dangerous because most people are not reading the details and they just enter an address even though they cant receive funds on one of the networks and we are then risking losing funds that cannot be recovered. This way they will be confronted with all chains.. 
What do you guys think?

alireza-sharifpour:
@maryjaf

maryjaf:
When I try to remove the address I get this error This field is required
as below pic 
<img width="744" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/de3e7611-a495-408e-a117-aa7146615423">
and by taping on close icon this address is removed, it seems we need the flow of removing address in new design 
@alireza-sharifpour @mosaeedi

maryjaf:
2-
plz check this scenarios 
-for a new project add a wallet address which is used in another project for example (0xde349606FB2B3ff792c23dD30dD3d39a94215320)
- this meesge is shown
This address is already used for another project. Please enter an address which is not currently associated with any other project.
-by tapping on close this address is removed from the box 
- but by taping on publish this message is shown despite

<img width="1223" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/c96bbb4e-c5be-43b7-9c9d-79c56aef73e4">

<img width="1338" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/5045564c-dbd7-42cc-8d64-4225c23bd5bf">

[...]

PR #2985, opened 2023-08-01
Update receiving addresses

alireza-sharifpour:
#2613


RamRamez (approved):
Thanks @alireza-sharifpour

PR #2990, opened 2023-08-06
Fix create project input

alireza-sharifpour:
#2613


MohammadPCh (approved):
LGTM, thanks @alireza-sharifpour

Issue #2774, opened 2023-06-15
Adding Goerli OP as a supported network to staging 


MoeNick:
it includes all subgraphs requests plus:

- change network modal 
- Create edit projects page
- My account manage recipient's address
- Verification flow: manage other addresses
- Tokens logo . token list in the donation page
- checking the balance in donation list to be correct

MoeNick:
cc: @laurenluz

laurenluz:
Thank you @MoeNick for creating this issue! do we need also to assigned someone for subgraph support?

MoeNick:
As @maryjaf said and I see, we don't have OP Goerli on staging. 
Im surprised how it gets closed directly from sprint planning.

I asked @RamRamez to review this PP and merge it to staging so we can test it.

MoeNick:
I meant this PR:
https://github.com/Giveth/giveth-dapps-v2/pull/2793

MoeNick:
can be tested on giveconomy branch

maryjaf:
1-
in verification form "optimism" is still shown , not optimism Goerli
![image.png](https://images.zenhubusercontent.com/62fe7a8b57c1a1770575ea40/6e53dbb2-4461-424c-ba9d-e1e045b951bf)

2-
I created a new project(testopgoerli) and set a OP Goerli wallet address for this project, when I try to add this address into other chain from manage wallet address in my account I get an error
![image.png](https://images.zenhubusercontent.com/62fe7a8b57c1a1770575ea40/2cc48bca-cbda-47f4-a32e-76045b26ff87)
![image.png](https://images.zenhubusercontent.com/62fe7a8b57c1a1770575ea40/58831c8b-6838-4ec7-b45e-49ccbc7a6859)

MoeNick:
@RamRamez can you take a look?

maryjaf:
3-
![image.png](https://images.zenhubusercontent.com/62fe7a8b57c1a1770575ea40/9130ddb0-895a-412e-a25f-3f1d1c53e08e)
The icon is not shown correctly in donation page

maryjaf:
Plz check these 3 problems @RamRamez

RamRamez:
> Plz check these 3 problems @RamRamez

@maryjaf I fixed 1 and 3 (it's waiting for review). I couldn't reproduce No 2. Can you test it again?

maryjaf:
> I couldn't reproduce No 2. Can you test it again?

[...]

PR #2991, opened 2023-08-06
Fix app to auto fetch networks

RamRamez:
#2774


MohammadPCh (approved):
LGTM, thanks @RamRamez

PR #3044, opened 2023-08-17
set studio subgraph for gnosis and mainnet on production

RamRamez:


MohammadPCh:
@RamRamez Thanks a lot, We did it in another PR.

Issue #2757, opened 2023-06-12
QF - post-round, show the actual matching funds sent to the project


laurenluz:
After a round ends, we will take the donation tx data, so some post-round sybil analysis... remove sybils... and then (manually?) recalculate the matching funds that should go to each project.

Then, we will send the actual matching funds to the projects from donation.eth via disperse.

For our MVP, these funds will be sent out on Optimism chain.

After this we will need:
- To show the amount of matching funds sent to each project in the frontend on the [donation tab](https://www.figma.com/file/C3HykACu0eNImhAqu0our8/Giveth-projects?type=design&node-id=2689-95212&t=BWuiQb1LBHFFrvVA-4)
![image.png](https://images.zenhubusercontent.com/5fdfac16a58523a866ddb68d/3df7e0b1-6b0f-4522-8ea6-74085a64b4e2)
- Nice to have: to trigger a notification when the matching funds are sent to alert project owners that they have gotten matching funds. (we could also just manually send out an email and cut this from scope)

Chatted w/ @CarlosQ96 about this. Looping in @mohammadranjbarz @MoeNick 

Related to #2694

divine-comedian:
@CarlosQ96 will jump to it this week 

found a bug in exporting donations to google spreadsheet where we apply filters for QF, it is filtering in the view but its not sending the filtered list to the spreadsheet.

laurenluz:
once we have the matching funds actual amounts, send spreadsheet to @CarlosQ96 who will manually input them in the database

laurenluz:
@divine-comedian when you send the funds, can you post the tx links here?

@CarlosQ96 - then please input them in the database.

@maryjaf - once that's done, please test to make sure that on production, the projects that were part of the QF alpha round have the correct "Matching funds" amount shown on their project page. You can see what the "Matching funds" amount should be for each project from [this spreadsheet](https://docs.google.com/spreadsheets/d/1UDg3YT59xaLdh8RDlFC0ttvKRCTumMBbsHMvFR6Dl1M/edit#gid=1175767477).

For example, on the [commons stack project](https://giveth.io/project/commons-stack?tab=donations), the matching funds amount should be 702.24 (not 450 as it shows now).

![image](https://github.com/Giveth/giveth-dapps-v2/assets/75490971/d754df3a-342c-452f-a333-9f2d38970a26)

divine-comedian:
pending tx to sign here:
[...]

PR #3064, opened 2023-08-21
changed estimated matching format

alireza-sharifpour:
#2757


MohammadPCh (approved):

Issue #3068, opened 2023-08-21
Wrong error message on editing the project


RamRamez:
In the edit project page, consider I'm not owner, app show "Only project owner can edit the project"

but when I switch to the owner account, I can see the edit project page but I still get this error
"Only project owner can edit the project"

![image](https://github.com/Giveth/giveth-dapps-v2/assets/73441857/875d50e5-7daa-4d3a-9ac6-0bfa26ab70c3)

RamRamez:
FYI @maryjaf

maryjaf:
plz check this @RamRamez 
By editing a project with another user, the below pic is shown
[ Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBcDVvIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--05c70b8df1dcbc31080d32ecea8ec5401d6d9758/image.png)

RamRamez:
@maryjaf I didn't change this part, and we had this before.
@divine-comedian if you think this message is not informative, I can change it.

Issue #3066, opened 2023-08-21
Disable edit project view for cancelled projects


RamRamez:
This project is cancelled but I can still see the edit project page

https://giveth.io/project/test-listing

![image](https://github.com/Giveth/giveth-dapps-v2/assets/73441857/d686481f-e16d-4bf9-abcb-35f50963bafb)

MoeNick:
@maryjaf can u pls confirm

maryjaf:
Thanks @RamRamez 
Now if the user has edit link for a cancelled project  the below pic is shown
[Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdmxvIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--d1bd16f3f4d2502b30eedad9c9cd8db8a0c78d8c/image.png)

Issue #3058, opened 2023-08-20
Redirect link for Project cancelation  in notification list 


maryjaf:
When My project is cancelled by admin user I receive a notification like below:
Your project [test999999](https://giveth.io//project/test999999) has been canceled by an admin. More information can be found in our [terms of service](https://giveth.io/tos) .
by tapping on name of project I see this message ,it's better the showing message in this page is revised , or the name of project is not clickable, what do you think? @divine-comedian 
<img width="1431" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/c70f6e4f-7af2-4ab7-98d5-19afc7935aae">

divine-comedian:
Yep good point! Maybe we can add to this text and provide a link for the user 

`Oops! This project is no longer available or not found! If you think this is a mistake reach out to us on Discord`[link to discord.giveth.io]`"

MoeNick:
@divine-comedian This is a general error that prevents users from seeing the canceled projects page, I suggest we won't make the link in the notification message if the notif is for canceled projects.

MoeNick:
- [] Removing the link to the project, if it's a canceled project notification. 
- [] Editing the copy of canceled project single view.

MoeNick:
Cool, based on @maryjaf 's comment I edited the scope:
- [ ] Removing the link to the project, if it's a canceled project notification. 
- [ ] Editing the copy of the canceled project single, if the owner opens his/her own project. 

As we discussed with @RamRamez and @mohammadranjbarz ,  1st one is a backend issue

maryjaf:
project owner view
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBaTVtIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--e9ead17c86d3c3801a8927f3ac88f290c81d2fad/image.png)

Project view:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBakptIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--898aa34d2b9fc3394ab480fc0a3e1ba07560ab95/image.png)
Is that ok? @divine-comedian

divine-comedian:
@RamRamez - can we remove the toast notification? I think it's confusing

[...]

PR #3070, opened 2023-08-21
Editing the copy of the canceled project in single project view and edit project

RamRamez:
#3058 #3066 #3067 #3068


alireza-sharifpour (approved):
Thanks @RamRamez

Issue #3054, opened 2023-08-19
Adding at least one wallet address should be required in the create project page


RamRamez:
Right now, we don't have any conditions in front-end to check if wallet addresses are empty. Backend checks it and returns the true error but we should have it in the front-end too.

![image](https://github.com/Giveth/giveth-dapps-v2/assets/73441857/2baa4c7b-c90b-4351-822f-1d0f01a606b3)

PR #3075, opened 2023-08-22
At least one wallet address should be required

RamRamez:
#3054


MohammadPCh (approved):
LGTM, Thanks @RamRamez

PR #3099, opened 2023-08-27
Update user and header to wagmi

alireza-sharifpour:
#3115


MohammadPCh (approved):

Issue #3087, opened 2023-08-24
Improving showing "Amount" and "USD Value" in donation tab


maryjaf:
It seems it's better that for low amounts for example less than 0.001 instead of showing 0 , <0.001 is shown , showing 0 amount and 0 $ in donation tab doesn't make sense

<img width="1422" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/0e358da4-e0ed-4eb2-8628-1aaf4e760abc">
@RamRamez @MohammadPCh @divine-comedian

MoeNick:
Awesome suggestion.

divine-comedian:
I think @RamRamez can check this one out

maryjaf:
Same comment 
https://github.com/Giveth/giveth-dapps-v2/issues/3071#issuecomment-1706164587

Issue #3071, opened 2023-08-21
amount column should be wider in "My Donations"


maryjaf:
**Describe the bug**
This issue has been resolved in project's donation but in "My Donations"  still exists 
**To Reproduce**
1. Go to 'my account'
2. Click on 'my donation'

**Expected behavior**
The scroll bar should be removed

**Screenshots**
<img width="1251" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/3c18c95f-7a7b-43c1-9cdf-a8e10bf67caf">

MoeNick:
@maryjaf pls confirm

maryjaf:
Now for small amount "<0.01 " is shown, is this ok or it should be better changed to <0.001? @MoeNick 
[Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBZ1ZwIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--08ebfbc15ebca26d2982d06bccb540069387692a/image.png)

PR #3105, opened 2023-08-27
Improving showing amount and usd value in donation table

RamRamez:
#3087 #3071


MohammadPCh (approved):
LGTM, Thanks @RamRamez

PR #3168, opened 2023-09-08
conditionally show GIVeconomy on navbar

divine-comedian:
Currently on the dapp when a user switchs to a network like polygon or Celo by default this shows them their Gnosis balance. this can be confusing since they don't actually own those tokens on their currently connected network. 

Furthermore, since there is no GIVeconomy on those networks everything in the drop down menu will return 0's - making it not a very useful thing to show to the user. 

This will create a condition that checks the connected chain and remove the GIV balance component from the header completely if the user is connected on a chain with no deployment of the GIVeconomy.


MohammadPCh (changes requested):
Thanks @divine-comedian 
I suggest to use this code instead:
```ts
const networkHasGIV =
		chainId && config.NETWORKS_CONFIG[chainId].GIV_TOKEN_ADDRESS;
```

MohammadPCh (commented):
LGTM, Thanks @divine-comedian

MohammadPCh (approved):
LGTM, Thanks @divine-comedian

PR #3177, opened 2023-09-13
add toggles for selecting qf round

divine-comedian:
Related to https://github.com/Giveth/giveth-dapps-v2/issues/2915

### DIFF
--- a/src/components/views/project/projectDonations/QfRoundSelector.tsx
+++ b/src/components/views/project/projectDonations/QfRoundSelector.tsx
@@ -15,7 +15,6 @@ import 'swiper/css';
 import 'swiper/css/navigation';
 import { Flex } from '@/components/styled-components/Flex';
 import { IQFRound } from '@/apollo/types/types';
-
 interface IQfRoundSelectorProps {
 	selectedQF: IQFRound | null;
 	setSelectedQF: Dispatch<SetStateAction<IQFRound | null>>;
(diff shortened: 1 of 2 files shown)


src/components/views/project/projectDonations/QfRoundSelector.tsx
```diff
@@ -30,48 +29,70 @@ export const QfRoundSelector: FC<IQfRoundSelectorProps> = ({
 	const navigationPrevRef = useRef(null);
 	const navigationNextRef = useRef(null);
 
+	const sortedRounds = projectData?.qfRounds
+		? projectData.qfRounds
+				.filter(round => round.isActive)
+				.concat(projectData.qfRounds.filter(round => !round.isActive))
+		: [];
+
```
MohammadPCh, line 37:
```suggestion
	const sortedRounds = projectData?.qfRounds?.sort((a, b) => b.isActive - a.isActive) || [];
```

MohammadPCh (changes requested):
Thanks @divine-comedian , I left a suggestion for sorting and for adding navigation, can we use what we have in `SwiperPagination.tsx`?
please check `CampaignsBlock.tsx` to see how you can use it.

src/components/views/project/projectDonations/QfRoundSelector.tsx
divine-comedian: This actually doesn't work because the variables to use for sorting need to be strings or numbers - it can't do sorting operations on booleans - are you sure this is the best way?

divine-comedian:
> Thanks @divine-comedian , I left a suggestion for sorting and for adding navigation, can we use what we have in `SwiperPagination.tsx`? please check `CampaignsBlock.tsx` to see how you can use it.

using the pagination would increase the vertical height of the element which I don't think we need 

we also don't need to toggle an entire page scroll, each view can hold about 8 QF rounds and we've only had 1 so far, so we would need to have 16 or more QF rounds for this to make sense...

MohammadPCh (approved):
Thanks @divine-comedian

Issue #3081, opened 2023-08-23
Actions for Project with "Draft" or "Deactivated" status


maryjaf:
I think it's better that the **"Project Actions"** for **Draft** and **Active** project be different
For example now I can complete verification form for draft project 
Is this Ok? @divine-comedian @MoeNick 

1. Go to 'my projects'
2. Click on 'a draft project'
3. Click on 'Project Actions'
4. See 'Verify Project'
5. Complete verification form

**Expected behavior**

**Screenshots**
<img width="1078" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/f70c6aaa-42f0-4a80-9ecf-a9ac7717d8da">

and one more thing
I can activate  draft project by taping on **active project** on "project actions" with out needed to taping on **continue creation**,  is this accepted?  
<img width="1430" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/2e8266c3-0adb-420c-9d50-b437e81256ad">

divine-comedian:
Nice catch MJ! I'm wondering then what would be the appropriate "Project Actions" we should include when a project is in draft state - should we include any at all in this view? Maybe it's more confusing for the user.

maryjaf:
Yeah, I think the only action required for a project in draft status is to complete the project publish operation.
Maybe it is better to customize these project actions based on the status of the project  Active,Deactivate and Draft

divine-comedian:
Okay so what do you think:

## Active 
- All regular options 

## Deactivated 
- Edit Project 
- Activate Project

## Draft  
- remove actions drop down menu entirely

maryjaf:
> Okay so what do you think:
> 
> ## Active
> * All regular options
> 
> ## Deactivated
> * Edit Project
> * Activate Project
> 
> ## Draft
> * remove actions drop down menu entirely

Yes, exactly
But i'm not sure by removing actions for Draft project, how does the design be changed? 
For example the  "Continue creation" can be shown instead of project actions 
<img width="1432" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/db5e9b9d-a29c-4656-b404-34967d5f69bc">

divine-comedian:
- I think it's okay to have edit for a deactivated project - There is a case that a project owner might want to edit their project before reactivating it so its fine I think 

[...]

Issue #3055, opened 2023-08-20
Project Owner View: iPhone Mobile view issues


maryjaf:
**3- Owner view**

![image](https://github.com/Giveth/giveth-dapps-v2/assets/111529185/82d3a782-678c-49ce-926a-dbe5d465c453)

_Originally posted by @maryjaf in https://github.com/Giveth/giveth-dapps-v2/issues/2971#issuecomment-1651068002_

divine-comedian:
@alireza-sharifpour is there any update for this issue? Does it need more design?

mosaeedi:
The main difference on mobile is related to the **Actions button** and the **Boost button**.

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdEo2IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--136d321c08dd850ff89ae379b66d85556688b6e5/Screenshot%202023-09-07%20at%209.52.03%20PM.png)

The **Project Actions** button is now in the same card with the project details.
When user taps on it, the full-page menu will appear.

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdkY2IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--1830be2cab70dbcf9e08a147144338dc4e850efc/Screenshot%202023-09-07%20at%2010.04.08%20PM.png)


---

[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBdDE2IiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--d8523a9a98c379ce57e95fac54e881411678dc6f/Screenshot%202023-09-07%20at%209.55.12%20PM.png)

And the Boost button is on the same card as the donations detail,
When user taps on it, a full-page modal will appear to show boost details, and then they can go to the boost modal.

[...]

PR #3181, opened 2023-09-17
fixed project admin's active options on mobile

alireza-sharifpour:
#3055 #3081


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #2544, opened 2023-04-23
Public view | project tags


MoeNick:
**Please note:**

1. We can only have up to 5 tags on every project.
2. If a project has no tags, the place for tags needs to be empty. 

![Group 618.png](https://images.zenhubusercontent.com/61dd85b9a1419a0fc83f15d1/dff41652-1ff8-4959-8b84-474bc252c4a3)

## Responsive resource

**WIP.**

## Check the design
[Link to Figma](https://www.figma.com/file/C3HykACu0eNImhAqu0our8/Giveth-projects?node-id=134-72524&t=yMkADlANl4In8UtA-4)

MoeNick:
@maryjaf you can test.

maryjaf:
<img width="1436" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/f4c10c93-d80d-4ea4-ac34-56fdd65d51a7">
Now I see only verified tag for projects , should the campaign name be  shown? @MohammadPCh 
I added this project to a campaign but i didn't see the tag related to campaign 
<img width="1436" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/4ca8b085-ea07-4796-a72b-8ddb3306203b">

MohammadPCh:
@alireza-sharifpour could you please take a look?

alireza-sharifpour:
We don't have the Backend for this feature yet.

MoeNick:
@mohammadranjbarz is it easy to go or it needs efforts?

mohammadranjbarz:
> @mohammadranjbarz is it easy to go or it needs efforts?

No we need to implement something for that, I created an issue for that

https://github.com/Giveth/impact-graph/issues/1051

laurenluz:
Can we just skip the tags for the QF MVP... include only two options: "Verified" and "< QF round name >" ??

MoeNick:
we don't need campaigns, it's ok .

MoeNick:
Ok, after the QF release I think we can put it in backlog again

MoeNick:
I think this is now unblocked for front end ? Did you test the backed @maryjaf ?

MoeNick:
@maryjaf jan, seems this one is ready for testing too.

maryjaf:
1-project owner view
[...]

PR #3185, opened 2023-09-18
Campaign badges

alireza-sharifpour:
#2544


MohammadPCh (approved):
@alireza-sharifpour  LGTM, Thanks.

PR #3229, opened 2023-09-24
Gnosis safe multisig authentication support

mateodaza:


MohammadPCh:
@jainkrati do you have time to review this?

mateodaza:
@MohammadPCh  @jainkrati  
@divine-comedian approved from user perspective so we only need code review

I'll be off starting today until Nov 10th - hope you guys can send it live without me, I'll be hard to reach but available if any emergency 💜

divine-comedian:
Before you go @mateodaza can you leave some more info for the devs on what is happening to make this feature work? 

Then we might not need to bug you as much in order to hand this off and get it live.

mateodaza:
Sure! I'll explain how to play with it and the technical part

So pretty much, you need to go to https://app.safe.global/ and on  the custom app you set your localhost:3000

image

There, automatically the app will connect to your safe multisig, when you try to sign for authentication the following flow will happen

- You will be told that all signers on the multisig needs to sign and approve the tx for you to have the token
- You'll be requested to sign again from your personal wallet and sign, after this 3 scenarios may occur:

No session for the multisig
- Another modal will appear to set an expiration time (3 days, 1 week, 1 month) and create a new session
- Modals will explain all the process, but then the user must sign the first time so it creates a new session
- Nothing will happen until all signatures are validated

Pending session (missing signatures)
- A modal will tell you all owners need to sign and it's still pending, come back later

Active session
- On going session, you'll get a token and will be able to use it for managing the profile, stake, unstake, and whatnot

For this 3 sessions we removed the ability to log out, so it doesn't allow any confusions for the user. Everytime you're out of the authentication modal, you'll be reconnected to the gnosis safe app

...

The technical side pretty much summarizes in
- relevant files: SignWithWalletModal and user.thunks 
- I put a condition to make this flow occur only if the user is connected to a gnosis safe
- web3modal and wagmi made this easy to set the change of connectors when needed
- code should be self explanatory connecting to the backend


@divine-comedian @MohammadPCh @jainkrati

jainkrati:
I am taking a look today. Last few days have been packed. Will wrap this review today.

next.config.js
```diff
@@ -62,6 +62,25 @@ const moduleExports = withBundleAnalyzer({
...
+		return [
+			{
+				// Adding CORS headers for /manifest.json
+				source: '/manifest.json',
+				headers: [
+					{
+						key: 'Access-Control-Allow-Origin',
+						value: 'https://app.safe.global',
+					},
+					{ key: 'Access-Control-Allow-Methods', value: 'GET' },
```
jainkrati, line 75: dont we need other methods support too ? GET,DELETE,PATCH,POST,PUT

src/components/menu/UserItems.tsx
```diff
@@ -98,18 +100,20 @@ export const UserItems: FC<IUserItemsProps> = ({
...
-				}}
-				theme={theme}
-			>
-				<GLink size='Big'>
-					{formatMessage({ id: 'label.sign_out' })}
-				</GLink>
-			</Item>
+			{!isSafeEnv && (
+				<Item
+					onClick={() => {
```
jainkrati, line 105: Could you please use useCallback to cache it during re-renders. Ref: https://react.dev/reference/react/useCallback#usecallback

jainkrati (changes requested):
Once changes are done, pls dm in Discord and I will take another look

next.config.js
mateodaza: Nope, it's only for the dev environment to work inside safe's interface - we could add those methods but it's not needed

divine-comedian:
@mateodaza - when can we expect an update on this?

src/components/menu/UserItems.tsx
mateodaza: I don't think a useCallback is necessary, but in words of chatGPT advice

`Using useCallback isn't necessary in this case because it's meant for memoizing functions to prevent unnecessary re-renders when passing functions to child components. Since useIsSafeEnvironment is concerned with updating state in response to changing connectors, useEffect is already the appropriate hook for the job. Adding useCallback would not offer any performance benefits and would only complicate the code without any clear advantage.`

https://github.com/Giveth/giveth-dapps-v2/blob/e10a2f13994c489fa6c04d657103e7b53900e39d/src/hooks/useSafeAutoConnect.tsx#L25-L34

let me know if we still wanna do the change

Issue #3035, opened 2023-08-15
QF - When the round is over, but matching funds have not yet been send, we should show "pending" on the donation tab


laurenluz:
Right now, the alpha round funds have not yet been sent... so this should not show a matching amount... it should show "pending"

On production:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBa2tHIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--4114e1f81174e4d39453dcd1d18da4367c62b113/image.png)

What I want:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBa29HIiwiZXhwIjpudWxsLCJwdXIiOiJibG9iX2lkIn19--1a4252670d967159af43123403be541c92c02c78/image.png)

@mosaeedi can you let me know if you're ok with the "pending..." thing I did in figma? if so, we can pass to the devs.

mosaeedi:
@laurenluz WDYT about the copy?

[Link to Figma](https://www.figma.com/file/C3HykACu0eNImhAqu0our8/Giveth-projects?type=design&node-id=4119-19247&mode=design&t=779HUnPv9ppR6gCW-4)

laurenluz:
I made a small change! Let's implement it like this:

Note @alireza-sharifpour - The text "Alpha Round" here is an example... it should change based on the round, i.e. this should actually be < QF round name > 

![image](https://github.com/Giveth/giveth-dapps-v2/assets/75490971/45a3f199-2167-49f4-b978-6eedb90069a9)

maryjaf:
‌Based on issue's description I think this new design is for round that it is ended 
But now I see the below pic for optimism round that it will be ended in 30 days later 
plz check this @alireza-sharifpour 
<img width="1440" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/1c316eae-53ca-47e5-9b04-ece8c40e8713">

laurenluz:
yep! @alireza-sharifpour - if the round "is Open/ Active" - here we should show the estimated matching funds (as before).

the only time we show this "pending" design is when the round "is Closed" and the matching funds have not been sent yet/manually inputted into our backend.

cc @MohammadPCh

alireza-sharifpour:
[...]

PR #3236, opened 2023-09-25
Pending matching funds

alireza-sharifpour:
#3035


MohammadPCh (commented):
@alireza-sharifpour Thanks.

Issue #3203, opened 2023-09-20
fetchUserByAddress on user profile page not recognizing auth token


mateodaza:
On the pages/user/[address].tsx file we are fetching the user data on the getServerSideProps but we can't grab the token on ssr as it's being saved on the localstorage, this is raising a problem where we can't get authenticated/restricted data for the user even when it's signed

![image](https://github.com/Giveth/giveth-dapps-v2/assets/18689480/7e446d6f-b7c1-4f64-9c36-5a8113ca561e)

Two solutions for this may be
- Move from local storage to cookies so we can grab it on the ssr (HARD)
- Move this query to the client, or at least partially, prefetching all data and then doing it again on the client if there's a token

I'd go with second option

PR #3257, opened 2023-09-29
set myAccount true if the wallet address and singed user are same

MohammadPCh:
- #3203


alireza-sharifpour (approved):
Thanks @MohammadPCh

PR #3262, opened 2023-10-01
Specific chain qf fixes

### LINKED ISSUE #3178: QF - Only donations made on specified chains are eligible for matching
Polygon wants to run a QF round with Giveth where only donations made on Polygon are eligible for matching.
In the future, we may want to run rounds for only Optimism, only Gnosis, only Celo, etc.
We need to make some small modifications to our design to inform the user which chains are eligible for matching.

My suggestions:

1. Project Single Page - Add text under estimated matching that says "Only donations made on < Polygon > are eligible to be matched."

We should be able to specify which chains are eligible in admin bro. If all chains are eligible, there is no message. If two chains are eligible, we can say, for example " "Only donations made on < Polygon > and < Gnosis > are eligible to be matched."

image

2. Project Single Page - Donations tab - QF Round tab - We may want to show here ONLY the project recipient addresses for the eligible chains. We should still, however, show all the addresses on the "All donations" tab.

image

3. Donate page - If the user is connected to a matching-eligible chain - we show the page as normal for QF projects
image

4. Donate page - if the user is connected to a chain that is not eligible for matching in the QF round, we show the normal donate page (for non QF projects), but with an extra message (where the red box is in my image) that says "Make your donation eligible for matching. Switch to < Polygon >. *Switch Network*"

image

Anything I'm missing @MoeNick @divine-comedian @mosaeedi

alireza-sharifpour:
#3178

### DIFF
--- a/src/components/views/donate/CryptoDonation.tsx
+++ b/src/components/views/donate/CryptoDonation.tsx
@@ -95,7 +95,6 @@ const CryptoDonation: FC = () => {
 	const [erc20List, setErc20List] = useState<IProjectAcceptedToken[]>();
 	const [erc20OriginalList, setErc20OriginalList] = useState<any>();
 	const [anonymous, setAnonymous] = useState<boolean>(false);
-	// const [selectLoading, setSelectLoading] = useState(false);
 	const [amountError, setAmountError] = useState<boolean>(false);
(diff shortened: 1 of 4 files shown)


MohammadPCh (approved):

Issue #3281, opened 2023-10-03
Frontend for QF improved sorting


laurenluz:
backend at #2942 

On giveth.io/qf, the default sort should still be by "GIVpower", but we should add a new sort option to be "Amount Raised in QF". When the used sorts by "Amount Raised in QF" they should see a kind of sort + filter... where it shows only projects in the active QF round... and it sorts them based on the amount they have raised so far in the round.

@MohammadPCh - can you help with this?

or @alireza-sharifpour ?

cc @MoeNick

CarlosQ96:
Backend PR merged whoever takes this issue to add the filter in the frontend. Must use the following sortBy option:
```
  ActiveQfRoundRaisedFunds = 'ActiveQfRoundRaisedFunds',
```

maryjaf:
sorting QF projects based on Amount raised is buggy, plz check the below screen record @alireza-sharifpour 

[Uploading... Screen Recording 2023-10-05 at 3.06.20 PM.mov]

alireza-sharifpour:
I think we need an Icon for this new sorting option. @laurenluz @mosaeedi

maryjaf:
There is a problem that is not necessarily related to this new sorting, but it may cause the user to make mistakes and get confused
I am commenting here, if necessary, we can register a separate issue
There is already a **"Amount Raised"** sort that displays the total amount of project donation in order, when we have an active round, by selecting this field, we may see a view like the below pic , which will make the user get confused.
The first project has a higher total donation and therefore it is displayed in position one, but in the project card where we show the donation amount for this round, it is zero.
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBNm9LQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--96a369da65a88142deda92660b016c0c64f08ea3/image.png)

What we can do for this?
It should remains like before or should be changed @laurenluz

laurenluz:
I think we should change text on the original "Amount Raised" to be "Amount Raised All Time"

So we have 2 separate filters:
- Amount Raised All Time
- Amount Raised in QF

maryjaf:
> * Amount Raised All Time

@MohammadPCh @alireza-sharifpour

mosaeedi:
[...]

PR #3290, opened 2023-10-05
Add new sorting option

alireza-sharifpour:
#3281


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #3336, opened 2023-10-12
ETC on DApp | Sign-in and Switch network


MoeNick:

maryjaf:
[Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOFpCQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--fd215354a981d693f58a8c5636b55dc5b0990096/image.png)

maryjaf:
I got this error when I try to switch network @alireza-sharifpour

![image](https://github.com/Giveth/giveth-dapps-v2/assets/111529185/dc9d66de-9ea4-4197-99be-9216cfeb96da)

maryjaf:
In **switch network** modal the ETC network icon isn't shown for this project although this project supports Celo and ETC 

<img width="1284" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/05351ccc-0c53-4717-a37f-eb67b4e0ec8c">

PR #3349, opened 2023-10-15
Change network and sign in with etc

alireza-sharifpour:
#3336


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #3338, opened 2023-10-12
ETC on DApp| Create/Edit Project | add network


MoeNick:

maryjaf:
I got this error when I try to switch network @alireza-sharifpour 

<img width="1430" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/ee4d454a-ee16-4a31-b182-a952a0fb764d">

PR #3350, opened 2023-10-15
Etc create project

alireza-sharifpour:
#3338


MohammadPCh (approved):
Thanks @alireza-sharifpour

Issue #3363, opened 2023-10-18
By tapping on "Create a Project" button, nothing happens 


maryjaf:
When I am on a below page I see "Create A Project" button but by tapping on it ,nothing happens 

<img width="1024" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/9526e880-4dce-42c9-9e1e-7efc616b7559">

Screen record:

https://github.com/Giveth/giveth-dapps-v2/assets/111529185/6fb57466-aafe-4ea5-957a-cf2e3dc78846

cc: @MohammadPCh @divine-comedian

divine-comedian:
Ah very interesting - looks like nothing happens because the success page for project creation and "create project"  are on the same route - `/create`

I don't think this is an urgent case and we can have it in the p backlog

RamRamez:
@divine-comedian I moved the success view to /success route, so after project is created, user will be redirected to /success/[slug]

maryjaf:
Tthanks @RamRamez this issue has been resolved

PR #3376, opened 2023-10-22
Move success view to new route

RamRamez:
#3363


pages/success/[slug].tsx
```diff
@@ -0,0 +1,43 @@
...
+			props: {
+				project: data.projectBySlug,
+			},
+		};
+	} catch (error) {
+		return {
+			props: {},
+		};
+	}
+}
```
Comment (MohammadPCh) [inline pages/success/[slug].tsx:41]: @RamRamez Thanks, we don't need to build this page server side.

MohammadPCh (changes requested):

pages/success/[slug].tsx
Comment (RamRamez) [reply pages/success/[slug].tsx:41]: @MohammadPCh Do you mean removing getServerSideProps and fetching PROJECT_BY_SLUG client side?

Comment (MohammadPCh) [reply pages/success/[slug].tsx:41]: Yes, Thank you @RamRamez

Comment (RamRamez) [reply pages/success/[slug].tsx]: Done!

MohammadPCh (approved):
LGTM Thanks @RamRamez

PR #3388, opened 2023-10-25
Add .yarnrc.yml to specify which yarn version

kristoferlund:
Ensures right version for Giveth, simplifies for new developers that have yarn 2+ as their default version.


MohammadPCh (approved):
LGTM, Thanks @kristoferlund

PR #3416, opened 2023-11-08
remove uppercase transform from symbol selector

### LINKED ISSUE #3355: Donating USDC on Optimism for Giveth QF
**Describe the bug**
I had $44 worth of USDC on OP and 1) it would not populate and show that I had any in Giveth platform, even though it would show in my MetaMask Wallet. Then 2) when I tried to donate it would say I didn't have enough funds 

**To Reproduce**
1. Go to any OP QF round project and try to donate USDC on OP chain

**Expected behavior**
I expected to be able to use my $44 worth of USDC on OP to donate to the OP QF round on Giveth's platform 

**Screenshots**
image

image

**Desktop (please complete the following information):**
 - Browser: Google Chrome
 - Version: windows 11 version 22h2

**Additional context**

divine-comedian:
fix #3355


MohammadPCh (changes requested):
Thanks @divine-comedian 
But I think it may cause some problems in our code. and we should check all the tokens logo again. so I suggest to change the name to upper case.
wdyt?

divine-comedian:
@MohammadPCh I think its very nice if the tokens can show up with their proper capitalization. 

wxDAI, USDC.e agEUR - there are lots of instances of these capitalization differences 

Do you know specifically where we might encounter errors from not having these token symbols capitlaized?

MohammadPCh:
@divine-comedian, we display the tokens with their correct capitalization in the UI. However, for the logo files, we have converted all filenames to uppercase and also search for them in uppercase to ensure case sensitivity.

MohammadPCh (approved):
LGTM, Thanks @divine-comedian

PR #3447, opened 2023-12-03
Feat/solana_connect

aminlatifi:
Ref #3400


MoeNick:
@jainkrati fyi to lead the PR reviews

MohammadPCh (approved):
LGTM, Thanks @aminlatifi 
I think we need to handle the GIVeconomy UX too.

MohammadPCh:
@aminlatifi also, maybe we need to hide switch network.

PR #3460, opened 2023-12-06
Multiple issues for multisig integration

mateodaza:


MohammadPCh (approved):
@mateodaza I checked the `waitForTransaction` as you mentioned, It's great.
just user `Address` type from `wagmi` instead of `0x${string}` type.

Issue #3449, opened 2023-12-04
Other-network APRs in the GIVfarm show up wrongly!


laurenluz:
when I'm connected to the right network, I see the APRs on that network correctly, but the APRs shown on "other network farms" are all wrong!!

See?

connected to Polygon
![image](https://github.com/Giveth/giveth-dapps-v2/assets/75490971/a14c9bd4-4f8c-40da-bf82-23114eb07505)

connected to Gnosis
![image](https://github.com/Giveth/giveth-dapps-v2/assets/75490971/b9757fd0-5b8d-4068-9104-4f2a110d997d)

connected to Optimism
![image](https://github.com/Giveth/giveth-dapps-v2/assets/75490971/b324df1b-6cfe-4f5a-b8bd-427be7de7f6f)

@jainkrati @MoeNick - this is a bug - can we budget some time for this fix?

MoeNick:
I would ask @MohammadPCh take a look to see its small or not

jainkrati:
can @RamRamez take this if he has experience on this module?

PR #3469, opened 2023-12-07
Fix staking cards APR

RamRamez:
#3449


mohammadranjbarz (approved):
Thanks @RamRamez  LGTM

Issue #3255, opened 2023-09-28
Conditional Verification modals on project owner view of project 


GriffGreen:
I was trying to onboard a project and i couldn't find the verification button! Why is it hidden in a drop down... soooooo bad... we want them to verify!

I literally thought they werent able to verify until they were listed. I'm actually embarrassed now.

<img width="412" alt="Screenshot 2023-09-28 at 10 40 22 AM" src="https://github.com/Giveth/giveth-dapps-v2/assets/14945613/9e1a81ae-353c-4f7c-a986-df9d103464c3">

Please add back the buttons for Edit and Verify!

I wonder if we should have them start verifying from the project splash page too? Right after they create the project.

Right now it is REALLY bad UX and i'm sure we have a huge drop of rate from projects creating their project and applying for verification.

divine-comedian:
Sure @mosaeedi can take a look and revert it back

mosaeedi:
<img width="1312" alt="Screenshot 2023-10-04 at 1 02 38 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/13708322/a1eb5e9d-c738-43e3-bd2c-3d854a1b0be8">

Right now, when a project is not verified yet, we show a message below the project banner to encourage them to verify their project.
We can just add a button here to make it easier for user to apply for verification. 

### It can be like this...
<img width="676" alt="Screenshot 2023-10-04 at 1 02 08 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/13708322/6e7a8b7d-fa73-4772-a707-dd3c8698b8ca">

mosaeedi:
Regarding the Edit button, we can still keep it under the project action dropdown to keep the page clean.

mosaeedi:
<img width="1372" alt="Screenshot 2023-10-04 at 1 07 52 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/13708322/c91e8270-8930-41f6-96a9-14fccf8554ab">

divine-comedian:
I like this! let me chase @GriffGreen to see if this satisfies his concerns.

divine-comedian:
Griff says it's great! Let's go for it! @mosaeedi

mosaeedi:
So, we will add only the **Verify Project** button to this card verify project card.

[...]

PR #3471, opened 2023-12-08
fix verification button conditions

RamRamez:
#3255


src/components/views/project/projectActionCard/AdminActions.tsx
```diff
@@ -61,7 +66,7 @@ export const AdminActions = () => {
 			type: OptionType.ITEM,
 			icon: <IconVerifiedBadge16 />,
 			cb: () => setShowVerificationModal(true),
-			disabled: verified,
+			disabled: isVerificationDisabled,
 		},
 		{
 			label: capitalizeAllWords(
```
aminlatifi, line 72: Let's create a variable for each option (e.g. `editProjectOption`, `verifyYourProject` ,...) and at line 102 create the `activeOptoins` by explicitly mentioning active options and not by index
Change this 
	const activeOptions = isActive ? options : [options[0], options[2]];
With this
	const activeOptions = isActive ? options : [editProjectOption, activeDeactiveOption];

aminlatifi (changes requested):

src/components/views/project/projectActionCard/AdminActions.tsx
aminlatifi: I looked at the `DropDown` component, the `disabled` property has `isHidden` behaviour in fact. Thus, it's better only set the property value per option based on the `isAcive` value. I would suggest even changing the `disabled` property name.

aminlatifi (changes requested):

RamRamez:
@MohammadPCh can you please review this PR? we need to merge it before multisig release.

aminlatifi (approved):

MohammadPCh (approved):

Issue #3482, opened 2023-12-12
Projects in QF page are not being sorted by GIVpower


laurenluz:
@mohammadranjbarz @MohammadPCh - what's happening here?

https://giveth.io/qf/all

<img width="1412" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBOFpVQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--203ca80c9942764ea35e313ed2a74a9d879c1627/image.png" alt="image.png" />

cc @maryjaf 

we should try to fix this before the QF round goes live next week!

maryjaf:
Thanks @laurenluz for reporting
I've been checking, when we change the the default sorting from  GIVpower for example to Rank and then again set GIVpower, the order of showing project will  be resolved , we should check the reason of this behavior @MohammadPCh @mohammadranjbarz 
<img width="1218" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/a144c06f-26fc-4a29-a5f3-592b9b2473e2">

maryjaf:
It seems this is the reason @MohammadPCh
<img width="1349" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/aa9dcac1-bb13-440d-82cd-4ff6548df56d">

maryjaf:
Thanks @MohammadPCh it has been resolved on https://giveth-dapps-v2-git-fix-qf-page-sort-givethio.vercel.app/

MoeNick:
@laurenluz @jainkrati I think it worth a hot fix asap

laurenluz:
agreed! @MoeNick @MohammadPCh

maryjaf:
> Thanks @MohammadPCh it has been resolved on https://giveth-dapps-v2-git-fix-qf-page-sort-givethio.vercel.app/ and now the default sorting in QF page is GIVpower(instantboosting)

I'v been checkin on stg and I see this, plz take a look
I set this filter "Raised amount in QF" but it seems this sorting isn't ok
<img width="1419" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/682413eb-2221-456b-a047-57e58666a278">

PR #3485, opened 2023-12-13
fix QF page's default sort

MohammadPCh:
- #3482


aminlatifi (approved):

aminlatifi:
@MohammadPCh Please update the issue with your approach to fix the issue if you haven't already communicated it in a call. I approved since the changes was small and intuitive

RamRamez (approved):

PR #3498, opened 2023-12-18
Solana Support Patch

aminlatifi:
Made Solana Burner wallet only enable for test env
Added Solana Solflare Wallet

#3400


RamRamez (approved):

MohammadPCh (changes requested):
@aminlatifi Is it possible to move all these wallets to the outside of the provider?
so we don't need the useMemo either.

MohammadPCh (approved):
LGTM, Thanks @aminlatifi

Issue #3506, opened 2023-12-19
BUG!! GLO Dollars give an estimated matching of 0!! OH NO!


GriffGreen:
<img width="689" alt="Screenshot 2023-12-19 at 1 16 03 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/14945613/564073f3-72b4-4b38-947f-dcba2ce30b10">

laurenluz:
@MohammadPCh @mohammadranjbarz is this frontend or backend? we should resolve ASAP!

cc @jainkrati @MoeNick

mohammadranjbarz:
<img width="1450" alt="Screenshot 1402-09-28 at 11 47 54 at night" src="https://github.com/Giveth/giveth-dapps-v2/assets/9850545/8d508dc2-86b7-4f35-93aa-80c81ad379e3">

@GriffGreen  @laurenluz  @MoeNick  @CarlosQ96 
It's working now, maybe it was some problem in their network (Cant getting info from backend and then cant calculate estimated matching)

laurenluz:
@mohammadranjbarz it is still not working for me!

![image](https://github.com/Giveth/giveth-dapps-v2/assets/75490971/465b38e5-401d-4460-bc83-0a1526d1f236)

MoeNick:
@RamRamez you can check it on this issue, pls update us here if you checked the issue.

RamRamez:
I found the problem, I will send a hotfix soon.

maryjaf:
Thanks @RamRamez this problem has been resolved on production

PR #3517, opened 2023-12-20
Fix estimated matching of 0

RamRamez:
#3506


MohammadPCh (approved):
Thanks @RamRamez, LGTM

PR #3523, opened 2023-12-21
Add solana to create project

RamRamez:
#3393 #3473

Issue #3393, opened 2023-10-25
As a Project Owner, I want to add a Solana address to receive donations to


divine-comedian:
A Project Owner should be able to simply add a solana address to their project page that defines the recipient address of donations made on Solana 
Should consider any changes in #3078 

![image](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/90e94f4b-f29e-491c-ac28-dc1119b7a7cb)

MoeNick:
For adding the Addresses on the front end we have 2 views that should work:
- My projects, manage addresses, add address
- Add / Edit project view.

aminlatifi:
@RamRamez To add a Solana address to a project note these:
- NetworkId: 0,
- ChainType: SOLANA
- Address capitalization must be preserved

RamRamez:
@aminlatifi This error is coming from "walletAddressIsValid" endpoint. I suggest replacing "Eth address" with "Address" to contain Solana addresses as well.

![image](https://github.com/Giveth/giveth-dapps-v2/assets/73441857/ddb21902-e988-4c44-8c94-275f577cb684)

aminlatifi:
> @aminlatifi This error is coming from "walletAddressIsValid" endpoint. I suggest replacing "Eth address" with "Address" to contain Solana addresses as well.
> 
[...]

PR #3533, opened 2023-12-26
Add Solana to my projects

RamRamez:
#3393

This PR is ready for review. I will keep this draft until PR #3523 is merged.


src/components/modals/ManageProjectAddresses/AddNewAddress.tsx
```diff
@@ -39,18 +41,25 @@ export const AddNewAddress: FC<IAddNewAddress> = ({
 		formState: { errors },
 		setError,
 	} = useForm<IAddressForm>({ mode: 'onSubmit', reValidateMode: 'onSubmit' });
+	const chainType =
+		'chainType' in selectedChain ? selectedChain.chainType : undefined;
```
mohammadranjbarz, line 45: How about replacing it with 
`const chainType = selectedChain.chainType` it would have same result

src/components/modals/ManageProjectAddresses/AddNewAddress.tsx
```diff
@@ -39,18 +41,25 @@ export const AddNewAddress: FC<IAddNewAddress> = ({
...
 		const { address } = formData;
-		console.log('ASD1', address);
 		try {
-			const _address = getAddress(address) as Address;
+			let _address: string;
+			if (chainType === ChainType.SOLANA) {
+				_address = address;
+			} else {
+				_address = getAddress(address);
+			}
```
mohammadranjbarz, line 56: And also we can replace these lines with this
`const _address =  chainType === ChainType.SOLANA ? address : getAddress(address)`

src/components/modals/ManageProjectAddresses/ManageProjectAddressesModal.tsx
```diff
@@ -16,68 +17,56 @@ import config from '@/configuration';
...
-			if (address.networkId) {
-				WalletAddr[address.networkId] = address;
-			}
-		}
-		setAddresses(Object.values(WalletAddr));
-	}, [project]);
+	const selectedChainType =
+		selectedChain && 'chainType' in selectedChain
+			? selectedChain.chainType
+			: undefined;
```
mohammadranjbarz, line 41: Like above
How about replacing it with
`const selectedChainType = selectedChain.chainType` it would have same result

src/lib/helpers.ts
```diff
@@ -185,10 +186,21 @@ export const smallFormatDate = (date: Date, locale?: string) => {
...
-export const suggestNewAddress = (addresses?: IWalletAddress[]) => {
+export const suggestNewAddress = (
+	addresses: IWalletAddress[],
+	chain: Chain | NonEVMChain,
+) => {
 	if (!addresses || addresses.length < 1) return '';
 	const isSame = compareAddressesArray(addresses.map(a => a.address));
 	if (isSame) {
+		// Don't suggest EVM addresses for Solana input
+		if ('chainType' in chain && chain.chainType === ChainType.SOLANA) {
```
mohammadranjbarz, line 197: Cant we change it to 
`if (chain.chainType === ChainType.SOLANA)` ?

mohammadranjbarz (changes requested):
Thanks @RamRamez 
I just added some comments for improvement of codes, other parts seem good

src/components/modals/ManageProjectAddresses/AddNewAddress.tsx
```diff
@@ -102,14 +111,19 @@ export const AddNewAddress: FC<IAddNewAddress> = ({
...
+		} else if (!isAddress(address)) {
 			setLoading(false);
-			return 'Invalid address';
+			return 'Invalid ETH address';
 		}
 		return true;
 	};
 
-	const chainName = getChainName(selectedWallet.networkId);
+	const chainName = getChainName(selectedChain.id, chainType);
```
aminlatifi, line 126: Use general wallet provider (userGeneralWallet) chainName instead.

src/components/modals/ManageProjectAddresses/ManageProjectAddressesModal.tsx
```diff
@@ -16,68 +17,56 @@ import config from '@/configuration';
...
 				<Content>
-					{selectedWallet ? (
+					{selectedChain ? (
 						<SublineBold>
 							{formatMessage(
 								{
 									id: 'label.chain_address',
 								},
 								{
 									chainName: getChainName(
```
aminlatifi, line 67: Likewise

src/components/modals/ManageProjectAddresses/AddNewAddress.tsx
```diff
@@ -65,26 +74,26 @@ export const AddNewAddress: FC<IAddNewAddress> = ({
...
-						_add.isRecipient = true;
-						_add.address = _address;
-					}
-				}
+				_adds.push({
+					address: _address,
+					isRecipient: true,
+					networkId: selectedChain.id,
+					chainType,
+				});
```
aminlatifi, line 96: I don't understand why this new logic works! Was there any issue with the older implementation? Because the new one is pushing a new item to the list while the previous one wasn't!

aminlatifi (changes requested):

src/components/modals/ManageProjectAddresses/AddNewAddress.tsx
aminlatifi: Oh, the name is `useAuthenticationWallet` in your branch still

src/components/modals/ManageProjectAddresses/AddNewAddress.tsx
RamRamez: The type of selectedChain is "Chain | NonEVMChain" and "chainType" does not exist on type "Chain" so if I use "const chainType = selectedChain.chainType", I get this error: "Property chainType does not exist on type Chain"

src/components/modals/ManageProjectAddresses/AddNewAddress.tsx
RamRamez: Done!

src/components/modals/ManageProjectAddresses/ManageProjectAddressesModal.tsx
RamRamez: just like above I can't use this suggestion because "chainType" is not defined in "Chain" and I get typescript error

src/lib/helpers.ts
RamRamez: Likewise

src/components/modals/ManageProjectAddresses/AddNewAddress.tsx
RamRamez: which branch has "userGeneralWallet"? I can't find it

src/components/modals/ManageProjectAddresses/AddNewAddress.tsx
RamRamez: older implementation was using an array of wallet addresses and mapping that array to create "AddNewAddress" instances. It was confusing because not all projects have addresses in all chains. It's more reasonable to map Chains instead of wallet addresses. like in create project page and manage funds in verification view, we use Chains to map, so it's better to use Chains here as well instead of mutating the original wallet addresses array, and then we'll have a consistent code along the app.

Besides that, with this new logic, I'm using the original addresses that we fetch from backend and I'm not mutating it, that is the best practices. In older implementation, we were using a useEffect to mutate original addresses array, that is not necessary . So, now, when user adds a new address to a new chain, I call backend to update project addresses then use "push" to update front-end states as well.

src/components/modals/ManageProjectAddresses/AddNewAddress.tsx
aminlatifi: It's not merged yet. The `useAuthenticationWallet` is renamed to `userGeneralWallet` in another, but this functionality is not touched. You can go with useAuthenticationWallet so far

aminlatifi (changes requested):
Had answered yesterday here
https://github.com/Giveth/giveth-dapps-v2/pull/3533#discussion_r1438875849

RamRamez:
@MohammadPCh @mohammadranjbarz Can you please review this PR? it's ready.

src/components/modals/ManageProjectAddresses/AddNewAddress.tsx
aminlatifi: cc @jainkrati

src/components/modals/ManageProjectAddresses/ManageProjectAddressesModal.tsx
mohammadranjbarz: Got it, thanks

mohammadranjbarz (approved):
Thanks @RamRamez  Sorry for late review

src/components/modals/ManageProjectAddresses/AddNewAddress.tsx
jainkrati: @RamRamez to use useGeneralWallet

src/components/modals/ManageProjectAddresses/ManageProjectAddressesModal.tsx
jainkrati: @RamRamez to use useGeneralWallet here as well

src/components/modals/ManageProjectAddresses/AddNewAddress.tsx
jainkrati: @MohammadPCh to weigh in here.

src/components/modals/ManageProjectAddresses/AddNewAddress.tsx
RamRamez: @aminlatifi it's a wrong idea to use "useGeneralWallet" for fetching chain name here, because useGeneralWallet returns the current chain name that user is on, but in this case, we need the chain name that user has chosen to add wallet address to.
If I use useGeneralWallet, we will have:
image

cc @jainkrati

src/components/modals/ManageProjectAddresses/ManageProjectAddressesModal.tsx
RamRamez: @aminlatifi We can't use "useGeneralWallet" since it's returning user wallet current chain. In ManageProjectAddressesModal, we are not going to show current chain that user is on, the aim is to set a wallet address for the chain that user has selected.
Here, I used "useGeneralWallet", then I've chosen to set an address for Solana, but as you see, it says Gnosis, because I'm on Gnosis network:
image

cc @jainkrati

RamRamez:
I resolved the conflicts.

src/components/modals/ManageProjectAddresses/AddNewAddress.tsx
MohammadPCh: Guys thank you for this conversation,
@RamRamez about the first part, I don't agree with you about the confusing part. these are two different approaches and each of them has its benefits. 
and if we gonna talk about the second part, I do not agree with you about the best practices. we have a state and update it when the user adds something.
@jainkrati

RamRamez: Thanks @MohammadPCh. I disagree with you about both parts.

1. Could you please explain the benefits of older implementation?
You added a useEffect that applies unnecessary rerender by updating "addresses" state. We don't need to add an empty field for all chains in addresses array because user may not fill them all. You could simply replace "addresses.map" with "CHAINS.map" and remove the big useEffect that causes rerenders and unnecessary computations.

2. Could you please explain the bugs you have found in my codes?

aminlatifi (approved):

aminlatifi:
Thanks @RamRamez

src/components/modals/ManageProjectAddresses/AddNewAddress.tsx
MohammadPCh: @RamRamez, thank you for your response. 
Please note that my comment pertains to the subject Amin brought up in the call – that we should avoid refactoring when we're under pressure to deliver – and I confirmed his statement.
Regarding your first question, the previous approach was straightforward. I'm not sure why you found it confusing. We maintained an array of all chains and displayed them in the modal. When a user added an address, we updated its corresponding element. It's quite close to your idea, and we didn't reject it.
Generally, I don't check all the lines, I just take a quick look at them and check best practices.
Please be aware that when a 'useEffect' only runs once, it doesn't have a detrimental impact on code performance.
For the second question, could you please revisit my comment? Why do you think that way? 😄 
I want to emphasize once again that this discussion revolves around avoiding unnecessary changes that make the PR larger and more challenging to identify potential issues in the future.
@jainkrati

RamRamez: Thanks @MohammadPCh . Only 100 lines has changed and I can say it's a small PR.
But, after two weeks this PR is not yet merged. @MohammadPCh I'm really confused, have you reviewed this PR? Is it on approved state or requested changes state?

MohammadPCh: @RamRamez, as I mentioned, I haven't reviewed this PR yet because it seems unnecessary, given that you've already received approval from both Amin and Mohammad. Krati asked me to share my opinion on your discussions with Amin, and I have provided my feedback.

MoeNick:
So can we a assume it's done?

src/components/modals/ManageProjectAddresses/AddNewAddress.tsx
jainkrati: For reducing risk of regression, lets revert the unnecessary changes @RamRamez

PR #3557, opened 2024-01-01
Add new categories to giveth for polygon qf round

RamRamez:
#3540

PR #3567, opened 2024-01-05
Polygon QF round release

RamRamez:
A cherry-pick from #3557, #3570 and #3569


MohammadPCh:
@RamRamez Thanks, since we released it yesterday we don't need this anymore.

PR #3574, opened 2024-01-08
Solana switch networks

alireza-sharifpour:
#3543


src/types/config.ts
```diff
@@ -134,6 +134,7 @@ export interface NetworkConfig extends Chain {
 	uniswapV2Subgraph?: string;
 	WETH_TOKEN_ADDRESS?: Address;
 	MERKLE_ADDRESS?: Address;
+	chainType?: ChainType;
```
aminlatifi, line 137: Let's make it mandatory! Wdyt?

src/components/views/donate/common.types.tsx
```diff
@@ -1,3 +1,10 @@
+import { ChainType } from '@/types/config';
+
 export interface ISwitchNetworkToast {
 	acceptedChains?: number[];
 }
+
+export interface INetworkIdWithChain {
+	chainType: ChainType;
+	networkId: number;
+}
```
aminlatifi, line 10: Just a suggestion to pass config easier.
```suggestion
export type INetworkIdWithChain = Pick<NetworkConfig, 'id' | 'chainType'>;
```
If we define NonEVMChain as below, we could pass them directly as well
```ts
export interface NonEVMChain
	extends Pick<NetworkConfig, 'id' | 'name' | 'chainType'> {
	adapterNetwork: WalletAdapterNetwork;
	blockExplorers: {
		default: {
			name: string;
			url: string;
		};
	};
}
```

src/components/views/donate/helpers.tsx
```diff
@@ -30,12 +32,20 @@ export const prepareTokenList = (tokens: IProjectAcceptedToken[]) => {
...
+	text: string,
+) => {
 	return networks.map((network, index) => {
-		const name = config.EVM_NETWORKS_CONFIG[network]?.name;
+		// Access the network name using networkId or chainType based on the chainType
+		const name =
+			network.chainType === ChainType.EVM
+				? config.EVM_NETWORKS_CONFIG[network.networkId]?.name
+				: config.NON_EVM_NETWORKS_CONFIG[network.chainType]?.name;
+
```
aminlatifi, line 45: We have `getChainName` which works for EVM chains. However, I think it would be awesome to refactor it because it's not readable enough and made by assuming network id of non evm chains are always zero! which is not a strict rule and later we may modify it to non-zero and cause issues.

src/components/views/donate/SwitchToAcceptedChain.tsx
```diff
@@ -7,18 +7,26 @@ import {
...
 	const { formatMessage } = useIntl();
 	const { chain } = useNetwork();
 	const chainId = chain?.id;
 	const { switchNetwork } = useSwitchNetwork();
 
-	if (!chainId || !acceptedChains || acceptedChains?.includes(chainId)) {
+	// Update the condition to check if the current chainId is in the list of acceptedChains
+	if (
+		!chainId ||
+		!acceptedChains ||
```
aminlatifi, line 23: For solana we must rely on chainType so far.

aminlatifi (changes requested):
Thanks @alireza-sharifpour 
Left some comments and suggestions.
Please when the final version gets ready, review all changes and remove `console.log` lines.

MohammadPCh (approved):

aminlatifi (changes requested):

aminlatifi (approved):

aminlatifi:
Thanks @alireza-sharifpour

Issue #3568, opened 2024-01-05
P1 - Update estimated matching cap for Polygon QF round - on production by Jan 13


laurenluz:
@MohammadPCh @mohammadranjbarz @CarlosQ96 @RamRamez we need to update the estimated matching cap on the dapp again for the Polygon round!

They want a max of 10k Matic per project... and they want to have a matching pool of 250k matic...

so it will be 4%

We need this live before the round starts, so by Jan 13!

cc @jainkrati

jainkrati:
@MohammadPCh is this done? Today is the last date.

PR #3575, opened 2024-01-08
change cap to 4 percent

MohammadPCh:
- #3568


RamRamez (approved):
Thanks @MohammadPCh

aminlatifi (approved):
Thanks @MohammadPCh 
Just do you want to push this change to production as well?

MohammadPCh:
@aminlatifi Thanks, we need it for Jan 13th. I will do if we don't have a release until that day.

Issue #3606, opened 2024-01-11
Investigate integration with apollo client retry


aminlatifi:
The user queries may fail due to unstable network conditions. We have already implemented a manual retry on saving donations, but there would be more robust solutions.

It's an option that is worth looking at
https://www.apollographql.com/docs/react/api/link/apollo-link-retry/

aminlatifi:
@jainkrati What's your idea?

aminlatifi:
@RamRamez

aminlatifi:
I made this https://github.com/Giveth/giveth-dapps-v2/pull/3613

jainkrati:
Good call @aminlatifi for this.

PR #3613, opened 2024-01-11
Integrate retry link

aminlatifi:
- https://github.com/Giveth/giveth-dapps-v2/issues/3606


MohammadPCh:
@aminlatifi, have you tested it? I attempted to block it with Chrome, but it didn't retry. I'm uncertain if this is the correct method to test because I know that the retry link only works for network errors, and I'm not sure if this is a network error.

MohammadPCh (approved):
@aminlatifi Thanks, LGTM

alireza-sharifpour (approved):
Thanks @aminlatifi

PR #3634, opened 2024-01-16
Hiding notification count and hover for notification bell

mateodaza:


CarlosQ96 (approved):

RamRamez (changes requested):
@mateodaza I can still see the bell icon in the header. can you please check it?

RamRamez (changes requested):
@mateodaza I think it's better to set an env variable to easily turn on and off the notification center.

mateodaza:
> @mateodaza I can still see the bell icon in the header. can you please check it?

You can still see it, you won't check the count number or hover - the idea is to only allow user to see the notifications on the /notifications route

mateodaza:
> @mateodaza I think it's better to set an env variable to easily turn on and off the notification center.

will add this :)

Issue #3391, opened 2023-10-25
As a Donor, I want to make a donation to a project on Solana


divine-comedian:
When a user is connected to the solana network we should allow them to make a donation to a project. we should only allow specific tokens on solana network including the SOL native token.

related to https://github.com/Giveth/impact-graph/issues/1175

- SOL 
- mSOL
- USDC
- USDT
- RAY
- bSOL
- AUDIO
- MANGO
- Coin98

Here are the token images for some of the new tokens (USDT and USDC can be the same)

![iconC98Circle 5b1a0f1e](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/ed48f788-39bd-40f9-b17d-a0544dbbb572)
![mango](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/8fcf7b77-506f-425e-b5eb-dc57d94d9d64)
![audio](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/fbeed673-ca56-4043-bb24-d715cce258ac)
![bsol](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/5905d55a-a29b-4678-9227-36e54498c423)
![raydium](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/b4438467-a09e-4edb-abb2-d868146dac7e)
![solanaLogoMark](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/a9b61c3d-1f93-4450-abed-453b73f92506)
![solanaLogoMark](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/4ab30c11-70af-4bfd-8d0a-87dffa1897d4)
![mSOL](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/a57fc0cd-78de-4369-94a9-7985b46e502e)
![mSOL](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/07b6c242-02b6-4d79-a5f8-ffde9ed4aa03)

MoeNick:
Is it ready and assigned to @CarlosQ96 already, what else you guys need pls lmk?

CarlosQ96:
I think to start with just the Native token and a stable token might be the best option. 

But in any case this might not be THAT different than how we import tokens. We should work on the list to create a migration and add them.

But **4 hours** would be my estimate. We can re-use token logic.

[...]

PR #3645, opened 2024-01-17
Add Solana tokens to donate page

RamRamez:
#3391


src/types/config.ts
```diff
@@ -192,9 +192,15 @@ interface MicroservicesConfig {
...
 	id: number;
+	particleChainId: number;
 	name: string;
 	chainType: ChainType;
 	adapterNetwork: WalletAdapterNetwork;
+	nativeCurrency: {
+		name: string;
+		symbol: string;
+		decimals: number;
+	};
```
aminlatifi, line 203: Do you know if we need to define them here? Actually the SOL token always has the id of Solana system program, look at this in `useCreateSolanaDonation.tsx`
```ts

		if (address !== SystemProgram.programId.toBase58()) {
			throw new Error(
				'Token address is not native token address - not supported yet!',
			);
		}
```

src/config/development.tsx
```diff
@@ -51,9 +51,11 @@ const CELO_NETWORK_NUMBER = 44787;
 const CLASSIC_NETWORK_NUMBER = 63;
 const SOLANA_NETWORK: NonEVMChain = {
 	id: 0,
+	particleChainId: 103,
```
alireza-sharifpour, line 54: I think the networkId will be better name since the backend also use this name, what do you think @RamRamez

src/components/views/donate/CryptoDonation.tsx
```diff
@@ -80,6 +80,8 @@ const CryptoDonation: FC = () => {
 		isConnected,
 		balance,
 	} = useGeneralWallet();
+	const { connection: solanaConnection } = useConnection();
+	const { publicKey } = useWallet();
```
aminlatifi, line 84: The address passed from `useGeneralWallet` has the same value, we don't need to define this one.

src/components/views/donate/CryptoDonation.tsx
```diff
@@ -266,19 +270,18 @@ const CryptoDonation: FC = () => {
...
 		}
 		// Native token balance is provided by the Web3Provider
 		const _selectedTokenSymbol = selectedToken.symbol.toUpperCase();
 		const nativeCurrency =
-			config.EVM_NETWORKS_CONFIG[networkId!]?.nativeCurrency;
+			config.NETWORKS_CONFIG[
+				!walletChainType || walletChainType == ChainType.EVM
+					? networkId
+					: walletChainType
+			]?.nativeCurrency;
```
aminlatifi, line 284: We can detect it's a solana native token by comparing the token address with the system program id, as mentioned in another comment. If that matched, useGeneralWallet has computed the balance.

aminlatifi (changes requested):

src/config/development.tsx
RamRamez: Done!

src/types/config.ts
RamRamez: I've used "nativeCurrency" in "CryptoDonation" L:283. I've removed below code from that component, so need "nativeCurrency" here.
```
if (walletChainType === ChainType.SOLANA) {
			return setSelectedTokenBalance(
				BigInt(Number(balance || 0) * LAMPORTS_PER_SOL),
			);
		}
```

My idea was to avoid adding chain related hard coded data as much as possible. If we use "SystemProgram.programId", first we need to check if it's Solana or not, and we should avoid adding these conditions or checking chains one by one. I think it's better to try to detect chain data (like logo, currency, chain name, ...) automatically without adding conditions for each chain. Because we may add different non EVM chains in future, then we have to add lots of different `if` for each chain to check "nativeCurrency" and other chain related data.

src/components/views/donate/CryptoDonation.tsx
RamRamez: Here I've used the balance fetched from "useGeneralWallet" as you said (L:287).
It's better to avoid detecting Solana here. We would better avoid adding "if"s for each chain. I think it's better to detect chains automatically without adding conditions. e.g., here, I've removed below Solana condition, and it's working as expected:
```
if (address !== SystemProgram.programId.toBase58()) {
			throw new Error(
				'Token address is not native token address - not supported yet!',
			);
		}
```

src/components/views/donate/CryptoDonation.tsx
RamRamez: so, what's your suggestion?
The "walletAddress" passed "useGeneralWallet" has the wallet address of the current user in string format, but I don't need the wallet address. I need the publicKey of the SPL token to fetch user token account related to user selected token to fetch its balance.
We need to check token balance before allowing user to donate.

src/providers/generalWalletProvider.tsx
```diff
@@ -314,6 +324,50 @@ export const GeneralWalletProvider: React.FC<{
...
+				senderTokenAccountAddress,
+				receiverTokenAccountAddress,
+				publicKey,
+				value,
+				[],
+				TOKEN_PROGRAM_ID,
+			),
+		);
+		return await solanaSendTransaction(transaction, solanaConnection);
+	};
```
aminlatifi, line 369: Please move this method to src/hooks/useCreateSolanaDonation.tsx since it's the only place it's used and it's not needed to cluter the `generalWalletProvider`

aminlatifi (changes requested):

src/providers/generalWalletProvider.tsx
RamRamez: Done!

src/components/views/donate/CryptoDonation.tsx
RamRamez: Done! I changed the code to use "useGeneralWallet" instead of "useWallet" but we still need to convert it.
The walletAddress returned from "useGeneralWallet" is in Base58 format but here I need it in PublicKey format.

aminlatifi (approved):
LGTM, thanks @RamRamez

alireza-sharifpour (approved):
Thanks @RamRamez

Issue #3647, opened 2024-01-18
Sign in loading is not working on Solana


alireza-sharifpour:
WE should add loading while signing in with Solana to Signin modal as we have it with EVM wallets.

Leilani432:
Hello @alireza-sharifpour i will suggest you contact Solana support like i did  a couple of days ago and got support after i opened a ticket. So i will suggest you do the same so you can talk to a [Solana Live Support](https://helpassistance.live/solana) on their general chain link for definite answers.
You can use the  ticket ID DR3486 as well for a faster response.

alireza-sharifpour:
For testing.
The issue was with EVM sign in when the user clicks on sign in button in sign in modal, we have a loading indicator on the button.
We didn't have it on Solana but I added it.
It's ready for test. @maryjaf

PR #3648, opened 2024-01-18
added loading to solana signin

alireza-sharifpour:
#3647


RamRamez (approved):
Thanks @alireza-sharifpour

Issue #3399, opened 2023-10-26
Notify Users to switch network if a feature is not available on Solana (or Ethereum)


divine-comedian:
This should include:
- Projects that don't accept funds on Solana
- Projects that don't accept funds on EVM chains
- Recurring Donations 
- GIVeconomy pages
- NFT | PFP
- Gitcoin passport

---

## figma
https://www.figma.com/file/vMiyVd1Ly5LjgSyZrChVV8/Giveth-users?type=design&node-id=992%3A142551&mode=design&t=6gBBVPBLcG3Z1cC7-1

MoeNick:
@jainkrati shall we move it to the product backlog and assign it to @aminlatifi ?

MoeNick:
This issue has been defined before, on Nov 28th I asked for an assignment.
CC: @jainkrati

mosaeedi:
<img width="450" alt="Screenshot 2024-01-01 at 6 38 43 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/13708322/6551d739-5717-4150-9f3f-b06035e2726d">

I think we have to think about a copy that can be logical for all situations or probably use different copies in the modal!

[Link to Figma](https://www.figma.com/file/vMiyVd1Ly5LjgSyZrChVV8/Giveth-users?type=design&node-id=1722-32872&mode=design&t=ilrJO5Cd0G1kZBgQ-4)

MoeNick:
@mosaeedi 
the cases are dfferent:
 -Projects that don't accept funds on Solana: we have this change network modal like below
 
![image](https://github.com/Giveth/giveth-dapps-v2/assets/80677512/3135e92f-d126-43f4-a3f8-1c1012068a06)

 but we need the Solana logo to load sign-in with Solana Splash page

- GIVeconomy pages: we have this design, 
![image](https://github.com/Giveth/giveth-dapps-v2/assets/80677512/b78cdaf2-2b90-43af-bc19-a3c76a3b7aec)
![image](https://github.com/Giveth/giveth-dapps-v2/assets/80677512/4eb298bc-5284-4abe-a9ab-dc5a17290965)

But we need all of these logos load sign-in with Ethereum splash. 

- Recurring donation:
I think you should help me for this case.

mosaeedi:
I think this one can solve the issue.

[...]

PR #3649, opened 2024-01-18
Fix donation switch network and TokenPicker and Givpower Solana issue

alireza-sharifpour:
#3399

### DIFF
--- a/src/components/views/donate/SwitchToAcceptedChain.tsx
+++ b/src/components/views/donate/SwitchToAcceptedChain.tsx
@@ -10,16 +10,45 @@ import {
 import { INetworkIdWithChain } from './common.types'; // Import the type
 import { useGeneralWallet } from '@/providers/generalWalletProvider';
 import { ChainType } from '@/types/config';
+import config from '@/configuration';
 
 const SwitchToAcceptedChain: FC<{ acceptedChains: INetworkIdWithChain[] }> = ({
 	acceptedChains,
 }) => {
 	const { formatMessage } = useIntl();
-	const { chain, walletChainType } = useGeneralWallet();
+	const {
+		chain,
+		walletChainType,
+		handleSignOutAndSignInWithSolana,
+		handleSingOutAndSignInWithEVM,
+	} = useGeneralWallet();
 
 	const networkId = (chain as Chain)?.id;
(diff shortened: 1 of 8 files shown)


src/config/development.tsx
```diff
@@ -51,6 +51,7 @@ const CELO_NETWORK_NUMBER = 44787;
 const CLASSIC_NETWORK_NUMBER = 63;
 const SOLANA_NETWORK: NonEVMChain = {
 	id: 0,
+	networkId: 103,
```
aminlatifi, line 54: I am suspicious about using `networkId`
What if in the future an EVM chain comes with the same value?!

src/components/views/donate/SwitchToAcceptedChain.tsx
```diff
@@ -10,16 +10,45 @@ import {
 import { INetworkIdWithChain } from './common.types'; // Import the type
 import { useGeneralWallet } from '@/providers/generalWalletProvider';
 import { ChainType } from '@/types/config';
+import config from '@/configuration';
 
 const SwitchToAcceptedChain: FC<{ acceptedChains: INetworkIdWithChain[] }> = ({
```
aminlatifi, line 15: INetworkIdWithChain has chainType as well. So, let's rely on that for switching to Solana.

aminlatifi (changes requested):

src/config/development.tsx
alireza-sharifpour: What is your suggestion? we did this based on the naming of the backend.
I think if some day a chain has added and we decided to support it on Giveth, then we can change the name.
What do you think?

aminlatifi: I mean, you don't have to add anything here at this stage with your intended changes. So please don't modify it so far and let it for other PRs

aminlatifi (changes requested):

alireza-sharifpour:
Thanks @aminlatifi,
I did the suggested changes and also fixed the bug we had in TokenPicker on this page in this PR.

aminlatifi (approved):

RamRamez (approved):
Thanks @alireza-sharifpour

PR #3680, opened 2024-01-25
Add Solana node url

RamRamez:


src/providers/solanaWalletProvider.tsx
```diff
@@ -20,6 +20,9 @@ interface IProviderProps {
 }
 const { SOLANA_CONFIG } = config;
 const solanaAdapter = SOLANA_CONFIG?.adapterNetwork;
+const productionSolanaNode = process.env.NEXT_PUBLIC_SOLANA_NODE_URL;
+const developSolanaNode = clusterApiUrl(solanaAdapter);
+const solanaNode = isProduction ? productionSolanaNode! : developSolanaNode;
```
aminlatifi, line 25: Why not make it so general and use that every time the NEXT_PUBLIC_SOLANA_NODE_URL is defined?
```suggestion
const solanaNode =
	process.env.NEXT_PUBLIC_SOLANA_NODE_URL || clusterApiUrl(solanaAdapter);
```

aminlatifi (changes requested):

src/providers/solanaWalletProvider.tsx
RamRamez: Done!

aminlatifi (approved):

PR #3681, opened 2024-01-25
fix msol token logo

RamRamez:


aminlatifi (approved):

aminlatifi (changes requested):
Is it possible to rename the MSOL.svg to mSOL.svg? because that will be safer.

aminlatifi (approved):

Issue #3720, opened 2024-02-04
Hide MM from wallet adaptor modal


MoeNick:
HAPPENING TO SOME USERS: 
Connecting to Solana via the snap in Metamask is not working on DApp, so we must remove the Metamask option from the wallet adaptor modal to avoid confusion.

jainkrati:
@alireza-sharifpour can you pls take this up today itself ? @MohammadPCh and @aminlatifi to review

aminlatifi:
@MoeNick What's the problem with this extension? It's working for me, I guess as expected.

MoeNick:
@maryjaf can you pls make sure whether we have bug on this or not?

maryjaf:
I had tested connecting to solana  with mm and it worked for me

alireza-sharifpour:
Shall we close this issue?
@jainkrati @MoeNick @aminlatifi

MoeNick:
I'll keep it in the backlog for investigation on some users, I will unassign you and we will test with Mitch
Thanks

PR #3721, opened 2024-02-04
remove solana metamask

alireza-sharifpour:
#3720


aminlatifi (changes requested):
As mentioned [here](https://github.com/Giveth/giveth-dapps-v2/issues/3720#issuecomment-1925878616), I think this adapter works as expected.
So, there would be no need for this change.

jainkrati:
@alireza-sharifpour can you pls abandon or close this PR if not needed.

MohammadPCh:
Let's close this pull request since it aligns with everyone's decision! 😄

Issue #3745, opened 2024-02-13
FE - draft donation


MohammadPCh:
we have to save donations before sending them to the network.

MoeNick:
@maryjaf it's shipped, but just to make sure you did the QA?

PR #3747, opened 2024-02-13
Add draft donation

MohammadPCh:
- #3745


jainkrati:
@MohammadPCh changed look good, just one thing: fields which can have null values but in code we are force unwrapping them with ! , pls ensure FE doesn’t crash if value is null. Good solution could be to add default value in such cases or throw error in UX and continue instead of crashing

jainkrati (approved):

alireza-sharifpour (approved):
LGTM @MohammadPCh

Issue #3759, opened 2024-02-14
Pass draftDonationId param in donation creation


aminlatifi:
CreateDonation mutation now accepts a new parameter `draftDonationId`. Passing this will handle situation the save donations has take long time and draft donation matcher has already created the donation by draft information

MoeNick:
@maryjaf did you test this one? shall I close?

maryjaf:
Draft donation has been tested on prod and there is no problem

PR #3770, opened 2024-02-18
Add draft donation id to create donation

MohammadPCh:
- #3759


src/hooks/useCreateEvmDonation.tsx
```diff
@@ -27,6 +27,7 @@ export const useCreateEvmDonation = () => {
 	const { chain } = useNetwork();
 	const chainId = chain?.id;
 	const isSafeEnv = useIsSafeEnvironment();
+	const draftDonationIdRef = useRef(0);
```
aminlatifi, line 30: Would it cause any issue if we save `draftDonationId` as part of `createDonationProps`? if no, it will make the code cleaner and needs less modification

aminlatifi (changes requested):

aminlatifi (approved):
Thanks @MohammadPCh 
Just please be sure everything will be fine after this change by testing it in practice.

alireza-sharifpour (approved):
LGTM @MohammadPCh

PR #3773, opened 2024-02-18
Update new round banner with new structure

MohammadPCh:
This is a fixed for renaming QFBanner to activeQFBanner since we need to have another QFBanner called ArchivedQFBanner,


alireza-sharifpour (approved):
LGTM @MohammadPCh

Issue #3283, opened 2023-10-04
New Component for projects in 'My Projects'


divine-comedian:
Project owners will claim their streamed tokens from the My Projects page. 

This is a totally new component to see projects from a list view. 

This new component includes:

- [x] Veirifcation status
- [x] Listed status 
- [x] active qf round eligibility
- [x] project actions menu
- [x] project activation status 
- [x] project title
- [x] project created date 
- [x] total raised number
- [x] claimable recurring donations on connected network

https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/GIVeconomy?type=design&node-id=23293-17833&mode=design&t=3NphLaAiiOTZ5y7Q-0
![My accounts](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/56013b28-6ac6-4b04-a3cc-bddb17e20b29)

alireza-sharifpour:
I have some questions:
1- we need all the badges for all conditions for this section because they are different from our previous design.
<img width="189" alt="Screenshot 2023-10-26 at 6 55 03 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/48887817/ab76ff14-b638-408f-b95b-730d2fc15217">

2-This typography is not from the design system.
<img width="1077" alt="Screenshot 2023-10-26 at 6 55 47 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/48887817/8730c30a-dac9-42f6-be12-f186d1a791bc">

3-We need the badge design for verified and unverified, here is just verified design.
<img width="155" alt="Screenshot 2023-10-26 at 6 56 39 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/48887817/2480c629-8fe8-4cba-a46d-d69f7feddf74">

4-We need design for all conditions in this section.
<img width="195" alt="Screenshot 2023-10-26 at 6 57 10 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/48887817/fd203ad6-ec9d-4788-8521-0913b31d8649">

5- And also for this section if it has other conditions:
<img width="151" alt="Screenshot 2023-10-26 at 6 57 37 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/48887817/445415c5-0984-480e-b064-21526c8fe167">

6- What does this arrow do? and what is behind the text, i cannot see.
[...]

PR #3783, opened 2024-02-22
Fix project component issues

alireza-sharifpour:
#3283


MohammadPCh (approved):
@alireza-sharifpour, Thanks, LGTM

PR #3831, opened 2024-03-07
Feat add socialmedia

alireza-sharifpour:


RamRamez (changes requested):
@alireza-sharifpour I think showing the social media in single project view is a little buggy. For website we show "@" then only showing the url after "/" and for Youtube we show two "@"s in that section. Also for Discord I think it's better not simplify the URL, I think it's better to show the full URL.

image

alireza-sharifpour:
> @alireza-sharifpour I think showing the social media in single project view is a little buggy. For website we show "@" then only showing the url after "/" and for Youtube we show two "@"s in that section. Also for Discord I think it's better not simplify the URL, I think it's better to show the full URL.
> 
> image

I Agree with you, because they are links it's better to show the complete link,
I updated the code and I only remove https from the link to make them shorter and more readble.
@RamRamez

RamRamez (approved):
Thanks @alireza-sharifpour

Issue #3093, opened 2023-08-25
View & understand recurring donations being sent to a Project


divine-comedian:
We'll need to show on the project page a list of the recurring donations currently being made to the project.

This information needs to be included on the table:

- [x] date the recurring donation was started
- [x] donor name
- [x] network 
- [x] amount being donated monthly
- [x] total donated to this project from this donor from recurring donation
- [x] USD value of total donated amount

On the scroll bar above the table we should add in a new button - Recurring Donations - clicking this will change the table view to show only the recurring donations.

![Project Single - Donations](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/eeb13fc3-170e-4fe1-b590-497938fc1039)

jainkrati:
@MohammadPCh @mohammadranjbarz can you share date of completion for this from FE and BE perspective.

MohammadPCh:
@jainkrati, the front-end requires approximately ~16 hours to implement this.

divine-comedian:
Thanks @MohammadPCh - Good question on the edge cases, I think a few things will need to be considered. 

## $ USD value of total recurring donations

First is this part -
![image](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/dc51be6d-6694-43e4-8507-7f77a58e7cb2)

This is the dollar value sum of the active recurring donations going to this project for every token (and at a later point, for every chain). Do you think we can do it?

On the QF donation tabs we should not add this text, it only applies to the all time donation & recurring donation tabs for the project page.

## Inactive donations

Other thing in the table itself:
![image](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/c9607f99-3078-468d-96b5-45a80bdc366a)

We should handle when a recurring donation is terminated by the user or ends on its own. Since technically it is no longer streaming funds we should reflect that on the table. The simplest change we can change the amount to "Finalized"

![image](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/beb509e1-d5a5-4946-96ff-e6a919cc70fd)

Also when the donation is no longer active will we have issues querying things like the total amount of tokens sent? And also the dollar value of that total amount. 

I updated https://github.com/Giveth/impact-graph/issues/1142 and also made https://github.com/Giveth/impact-graph/issues/1243



## tx link 

on the total donated column there is a link - while it is a bit complicated to show all of the recurring donation data we can show the  'createFlow' tx - or whatever batch tx that included the createFlow function call. 

[...]

PR #3837, opened 2024-03-11
Project recurring donation table

alireza-sharifpour:
#3093


MohammadPCh (approved):
Thank you, @alireza-sharifpour.
@jainkrati, I will merge it since we need it for the presentation

PR #3843, opened 2024-03-12
fix recurring donation hash

alireza-sharifpour:


MohammadPCh (approved):
Thank you, @alireza-sharifpour.
@jainkrati, I will merge it since we need it for the presentation

Issue #3441, opened 2023-11-29
Project Owner: Claim Recurring Donations from Anchor Contract


divine-comedian:
Related to #3094 

Once #3440 & #3439 are implemented we will need to claim recurring donations from the project's Anchor contract instead of their project recipient address, since the anchor contract will be the recipient of the super tokens via recurring donation. 

This starts by calling the execute() function in the Anchor contract by the projectOwner
https://github.com/allo-protocol/allo-v2/blob/main/contracts/core/Anchor.sol#L73C4-L73C4
```
function execute(address _target, uint256 _value, bytes memory _data) external returns (bytes memory)
```
We target the supertoken Contract and encode some callData to call this function: 
```
function downgradeTo(address to, uint256 amount) external virtual override
```
https://github.com/superfluid-finance/protocol-monorepo/blob/dev/packages/ethereum-contracts/contracts/superfluid/SuperToken.sol#L715C4-L715C5

This will allow us to downgrade the super tokens held by the anchor contract to regular tokens and give them to the project recipient address

# How to claim multiple tokens
while this is possible via the SDK we will need to figure out how to do this via encoded callData 

the superfluid docs only provided information on batching calls via SDK

this is an open research question @jainkrati 
cc: @alireza-sharifpour 

# Sequence 
![image](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/569ceb6e-17bd-4723-b788-af2554808859)

divine-comedian:
Loop through a list of multiple downgradeTo function calls - depending how many tokens - if we loop through multiple downgradeTo calls - if we go beyond the gas limit the tx reverts - more complicated to call the batch call

if you want to downgrade and transfer it will be more complex. 

what are the thigns we can change? allo protocol is from gitcoin 

what are the changes that gitcoin made? what can we change and what can't we? we can't change this execute - make a middle contract and change it multiple times? 

project owner is EOA - make a contract inbetween that handles batching the function 

middle contract is called by project owner which calls the anchor contract which executes the batch calls 

do we need to deploy two contracts for one project per contract? middle contract can be owned by Giveth? 

we have two options - one option is to have a smart contract that owns all the anchor contract - don't have to redeploy everytime - allow people to call the anchor contracts and downgrade the tokens

if it has some security issues we don't like we create one clone per project - the front-end will need to call one function - if we have to engineer somethign with the database w/ the front-end if its just a smart contract that will be there forever 

one token at a time 

[...]

PR #3852, opened 2024-03-14
Add refetch to claim balance

alireza-sharifpour:
#3441


MohammadPCh (approved):
@alireza-sharifpour LGTM

PR #3882, opened 2024-03-25
remove loginWallet resolver

RamRamez:
we are removing this from impact-graph

https://github.com/Giveth/impact-graph/issues/1316


alireza-sharifpour (approved):
Thanks @RamRamez

Issue #3439, opened 2023-11-29
Donor: Create Allo Profile in Recurring Donation


divine-comedian:
Extension of #3091 

There is an extra step involved when a project is going to receive its first recurring donation - We need to setup a profile for it on Allo Protocol - This deploys an AnchorContract which becomes the recipient of the recurring donation. 

cc: @MohammadPCh FYI

## System Flow
![image](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/12fb5b20-a2ef-410c-9dff-aa2900bfc41d)

This is an extra modal/transaction for the Donor, who pays the tx and sets up the profile on behalf of the project. This is done by calling `createProfile` on the registry contract

https://github.com/allo-protocol/allo-v2/blob/main/contracts/core/Registry.sol#L118

The Registry can be found here on OP Goerli:
https://goerli-optimism.etherscan.io/address/0x4AAcca72145e1dF2aeC137E1f3C5E3D75DB8b5f3

And on OP Mainnet: 
https://optimistic.etherscan.io/address/0x4AAcca72145e1dF2aeC137E1f3C5E3D75DB8b5f3

## Creating an Allo Protocol Profile 
We call this function in the registry contract: 
https://github.com/allo-protocol/allo-v2/blob/main/contracts/core/Registry.sol#L118

Which takes a few arguments:
```
   function createProfile(
        uint256 _nonce,
        string memory _name,
        Metadata memory _metadata,
        address _owner,
        address[] memory _members
    )
```
### nonce
- unique integer used with the owner address to create the address hash of the anchor contract 
- must use a different nonce for project owners who own multiple projects (can we use the projectId in our DB?)

### name
- can use the project slug or project name

### metadata
- empty for now

### _owner
- the address of the project owner 

### _members
- leave empty for now

## Sequence
Here is the flow for when a user starts a recurring donation from the UI:
![image](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/507367fb-f215-47ef-8531-04e91c131308)

## Figma
https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/GIVeconomy?type=design&node-id=23970%3A14336&mode=design&t=Y0WYEAUOBSeS2TK1-1

![image](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/9a54a058-0c94-4df8-ae59-0f569f37e3c1)

jainkrati:
[...]

PR #3887, opened 2024-03-26
Fix anchor issue donation modal

alireza-sharifpour:
#3439


MohammadPCh (approved):

Issue #3897, opened 2024-03-26
Update PDF of quadratic force on Giveth to the new version ASAP :-D


laurenluz:
@jainkrati - we have a new PDF that should be hosted now at https://giveth.io/QuadraticForce

can we get someone to do this asap? Jake needs it so he can start pushing our fundraise for the next QF round

it's here: 
[Giveth_Report_QF_2024 this one.pdf](https://github.com/Giveth/giveth-dapps-v2/files/14762647/Giveth_Report_QF_2024.this.one.pdf)

cc @SnakeGH

alireza-sharifpour:
I changed it and the PR is ready but the previous file was 1.1 MB and this one is 17.9 MB.
Is it OK or do we need to compress the file? @laurenluz

SnakeGH:
we compressed the previous one, so if the quality stays the same, please do so again

alireza-sharifpour:
> we compressed the previous one, so if the quality stays the same, please do so again

Thanks for letting me know, I compressed it and left a PR. It's ready for other devs to review and after that we will have it on production.

SnakeGH:
@jainkrati looks like one more edit is needed, as I see "OP Mainnet" instead of reading "Optimism Network" in the section of how much DAI was used for the round under the Giveth Optimism Round on page 3 
![Uploading Screenshot 2024-04-03 at 6.29.47 PM.png…]()

laurenluz:
@SnakeGH are there any other changes that will be needed before this goes to a dev? like do we want to add something in there about the extra $10k and we'll distribute your token? (provided enough liquidity)?

SnakeGH:
> @SnakeGH are there any other changes that will be needed before this goes to a dev? like do we want to add something in there about the extra $10k and we'll distribute your token? (provided enough liquidity)?

As the reigning Queen of QF @laurenluz ill follow your thoughts on this. Is it time to add the extra info about using their own token?

SnakeGH:
[Giveth_QF_Force_2024.pdf](https://github.com/Giveth/giveth-dapps-v2/files/14871729/Giveth_QF_Force_2024.pdf)


this is the most recent @jainkrati but if @laurenluz wants to add the $10K in tokens for distro stuff, then Rodri will need to add that in, as its not included right now

laurenluz:
[...]

PR #3908, opened 2024-03-27
update QF pdf file

alireza-sharifpour:
#3897


MohammadPCh (approved):

Issue #3909, opened 2024-03-27
Bug | Claim recurring donation on chains other than op


maryjaf:
**Describe the bug**
-My network is set on for example Gnosis,  the"claim recurring donation" is shown in project actions  and it is enable 
-By tapping on claim recurring donation in project action , the loading icon is shown 

I think It would be enable only on optimism, am I right?
or in the  modal a message be shown with "switch network to op" button  ?
cc: @alireza-sharifpour @divine-comedian

maryjaf:
https://github.com/Giveth/giveth-dapps-v2/assets/111529185/90621106-ebf6-4ee1-8c9d-1133b08664ba

divine-comedian:
Similar here @alireza-sharifpour - we should handle if the user tries to claim recurring donations while on the wrong network. 

If we require them to be on Optimism to fetch super token balances etc.. then when they click the claim button we should send the switch network request to their web wallet.

alireza-sharifpour:
It's ready for testing @maryjaf

maryjaf:
Thanks @alireza-sharifpour 
It has been resolved and by tapping on cliam recurring donation switch network modal to op is shown
<img width="1364" alt="image" src="https://github.com/Giveth/giveth-dapps-v2/assets/111529185/583a189e-93d8-476a-ac58-7d9a56d6bb55">

PR #3917, opened 2024-03-28
Claim switch network

alireza-sharifpour:
#3909


MohammadPCh (approved):
@alireza-sharifpour Thanks

Issue #3285, opened 2023-10-04
Donor: Deposit/Withdraw from Stream Balance


divine-comedian:
From #3092 the donor will have an option to either deposit more tokens or withdraw tokens from their "stream balance" - this is actually them simply upgrading/downgrading their tokens to super tokens & vice versa.

![image](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/8314c833-e30d-4687-a041-dc4077a1d45e)
https://www.figma.com/file/nVoinu0tgJ565enN5R4WDE/GIVeconomy?type=design&node-id=23293-17833&mode=design&t=7J6iMuXHXGGPhHxz-0

Users will have two options from this modal - Deposit (Upgrade) & Withdraw (Downgrade)

underneath we show:

- [ ] Stream Balance (balance of super tokens)
- [ ] outflow rate 
- [ ] # of projects receiving tokens from this account for this token

under this we have an input field for the user to select an amount of tokens to deposit or withdraw. 
below this input field we show a token balance which changes based on if they selected deposit or withdraw. 

- [ ] deposit - we show balance of regular tokens available to be deposited
- [ ] withdraw - show balance of super tokens available to be withdrawn 

## User Errors

- [ ] If a user has active streams for this token, user cannot withdraw tokens to bring expected date when stream balance runs out below 1 day
- [ ] user cannot deposit/withdraw more tokens than their balance
- [ ] user cannot proceed if amount entered is blank or 0

MohammadPCh:
@maryjaf to test and close

maryjaf:
The item 2 of below comment is also related to this issue @MohammadPCh 

https://github.com/Giveth/giveth-dapps-v2/issues/3286#issuecomment-1990961456

divine-comedian:
@maryjaf - yes this makes sense, you withdrew 40 tokens from your stream balance but you did not modify how much you are donating monthly to a given project, 

Since you are technically donating per second you can see that your stream balance will last less than one month now since 40 > 29.9

maryjaf:
> yes this makes sense, you withdrew 40 tokens from your stream balance but you did not modify how much you are donating monthly to a given project,
> 
> Since you are technically donating per second you can see that your stream balance will last less than one month now since 40 > 29.9

I meant that when it is shown that the total stream balance is 29 and on the other hand it is shown that the "total recurring donation" is 40, which is greater than 29, it may cause the user to make a mistake.
[...]

PR #3919, opened 2024-03-28
fix next start new donation link

alireza-sharifpour:
#3285


MohammadPCh (approved):
@alireza-sharifpour Thanks,

Issue #3904, opened 2024-03-27
Add confirmation toasts to deposit/withdraw


divine-comedian:
When a user is depositing tokens into their stream balance or withdrawing tokens from their stream balance we should show a toast notification as a confirmation when it is successful. 

We can add in a green confirmation modal that we use similarly in other transaction modals

![image](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/6298d6d6-70e8-413a-919c-2b16224fde97)

### For Deposits 
"Deposit to your Stream Balance was successful."

### For Withdrawals 
"Withdrawal from your Stream Balance was successful."

_Originally posted by @divine-comedian in https://github.com/Giveth/giveth-dapps-v2/issues/3286#issuecomment-1994532307_

Meriem-BM:
Hey @RamRamez can you tell me which page holds this logic, a link to it please.

RamRamez:
@divine-comedian do we need a design for this? or maybe we can simply show a fixed toast at the top of the screen?

divine-comedian:
We don't need any new design, please read and understand the issue. This relates to the modal for depositing/withdrawing from your stream balance in the recurring donations feature.  

We can use the existing toast components that we use on the donation confirmation flow.

Look for example how the deposit flow is currently - 

https://github.com/Giveth/giveth-dapps-v2/assets/67759413/9470b723-c84a-4671-a30d-b1b10262f596

When the tx is successfully we should have a toast just above the main button on the modal 

![Screenshot from 2024-04-16 07-53-44](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/eeb0f4f3-4f9c-4cb4-b5e6-934b5e0d0f12)

RamRamez:
Got it, thanks @divine-comedian

RamRamez:
@Meriem-BM please use this component as toast:
https://github.com/Giveth/giveth-dapps-v2/blob/37c18418f42501ccc5cd92503bfa626628747e0e/src/components/toasts/InlineToast.tsx#L40

and add that to this component:
https://github.com/Giveth/giveth-dapps-v2/blob/cd7104c9ca21984f44844c19ac51470069ec00e3/src/components/views/donate/RunOutInfo.tsx#L14

jainkrati:
@maryjaf pls verify and close

maryjaf:
Thanks @Meriem-BM this issue has been resolved 

[...]

PR #4027, opened 2024-04-18
fix: add success Toast upon a successful withdrawal

Meriem-BM:
Related to #3904


MohammadPCh (approved):
@Meriem-BM LGTM, Thanks

Issue #4042, opened 2024-04-22
Remove QF Related Sections from Donate Success Page


MohammadPCh:

MohammadPCh:
@maryjaf ready to test.

maryjaf:
Passport section is removed from success page for both one time and recurring donation
It shouldn't be removed from one time donation  success page 
![image](https://github.com/Giveth/giveth-dapps-v2/assets/111529185/a9b98096-526f-42dc-b8dd-af0ded8f4177)

maryjaf:
Thanks @MohammadPCh it has been resolved fixed 

recurring donation:
![image](https://github.com/Giveth/giveth-dapps-v2/assets/111529185/688bf370-2424-4312-b638-16ca78a1242e)

One time donation
![image](https://github.com/Giveth/giveth-dapps-v2/assets/111529185/097d4f67-1a5c-4592-b6f0-b7cb7af5e85d)

PR #4043, opened 2024-04-22
Remove qf section from success view

MohammadPCh:
- #4042

### DIFF
--- a/src/components/views/donate/DonateIndex.tsx
+++ b/src/components/views/donate/DonateIndex.tsx
@@ -40,7 +40,7 @@ import { DonatePageProjectDescription } from './DonatePageProjectDescription';
 const DonateIndex: FC = () => {
 	const { formatMessage } = useIntl();
 	const { isMobile } = useDetectDevice();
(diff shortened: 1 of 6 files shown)


src/components/views/donate/DonateIndex.tsx
```diff
@@ -40,7 +40,7 @@ import { DonatePageProjectDescription } from './DonatePageProjectDescription';
 const DonateIndex: FC = () => {
 	const { formatMessage } = useIntl();
 	const { isMobile } = useDetectDevice();
-	const { project, isSuccessDonation, hasActiveQFRound } = useDonateData();
+	const { project, successDonation, hasActiveQFRound } = useDonateData();
```
jainkrati, line 43: why was this renamed?

MohammadPCh: @jainkrati, it was renamed because the previous name, 'isSuccessDonation,' was incorrect as it suggested a boolean type, but it actually includes the success donation data.

jainkrati (approved):

PR #4046, opened 2024-04-22
Fix unauthorized problem

MohammadPCh:
Add logs to find the problem.

### DIFF
--- a/src/components/views/donate/RecurringDonationCard.tsx
+++ b/src/components/views/donate/RecurringDonationCard.tsx
@@ -229,6 +229,14 @@ export const RecurringDonationCard = () => {
 		'totalStreamPerSec',
 	);
 
+	const isFormInvalid =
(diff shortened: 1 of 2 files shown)


src/components/views/donate/RecurringDonationCard.tsx
```diff
@@ -224,10 +223,13 @@ export const RecurringDonationCard = () => {
 		}
 	}, [selectedToken, address, tokenStreams, project.anchorContracts]);
 
-	console.log(
```
jainkrati, line 227: why was this log removed?

MohammadPCh: @jainkrati, we removed it because it's now unnecessary. We initially added it to check our donation amount, but now we're confident everything is functioning correctly.

jainkrati (approved):

PR #4066, opened 2024-04-25
text & functionality edits

divine-comedian:
Fix some missing translations and copy improvements 

do some checks if the token is ETH or ETHx and add/remove visual components or add text warnings. 

remove warning on withdraw stream balance if user has no active streams 

prevent trailing zeroes on stream balance table

add warning text for when project owner is claiming ETH


MohammadPCh:
Thanks, @divine-comedian! I didn't check the functionality, but the code seems fine. I just asked a question about a part of the code that appears to be unnecessary.

divine-comedian:
> Thanks, @divine-comedian! I didn't check the functionality, but the code seems fine. I just asked a question about a part of the code that appears to be unnecessary.

I don't see your comment, perhaps you didn't submit it?

MohammadPCh (approved):
@divine-comedian Thanks

PR #4078, opened 2024-04-30
Fix: Recurring Donation Token Price

MohammadPCh:


src/components/views/donate/ModifySuperToken/ModifySection.tsx
```diff
@@ -132,7 +132,7 @@ export const ModifySection: FC<IModifySectionProps> = ({
 					{balance
 						? truncateToDecimalPlaces(
 								formatUnits(maxAmount, balance.decimals),
-								6,
+								balance.decimals / 3,
 							)
```
jainkrati, line 136: Curious why is this division by 3

jainkrati (approved):

PR #4102, opened 2024-05-01
[HOTFIX] improve verification before you start page

divine-comedian:
Add a link so projects can edit their project and add social media links from verification form.


MohammadPCh (approved):
Thanks @divine-comedian,LGTM

Issue #3910, opened 2024-03-27
Get contact information from projects in the verification form


laurenluz:
We should add new optional fields to the project verification form that allows us to DM them if needed on other platforms. 

In the project verification UI we will need fields for:

telegram
discord
twitter handle
whatsApp

@divine-comedian @WhyldWanderer anything else?

We need to collect this info so we can do better outreach around marketing initiatives like QF etc.

back-end related issue is https://github.com/Giveth/impact-graph/issues/1436

laurenluz:
I think this is at least a P2. Without having this contact info, the comms team & outreach folks end up doing way more working running around to get it, we've wasted a lot of our contributors' time because we don't have this readily available. This should be a "normal to do" thing... we need it.

laurenluz:
should we close this issus? I think it's redundant w/ the one in impact graph

divine-comedian:
No because this one is a front-end issue - we'll need the design space put into the verification form - the other one is related to saving information into the back-end & making it available in adminJS.

Tosinolawale:
![Project contact](https://github.com/Giveth/giveth-dapps-v2/assets/35389241/1d005c6c-29d1-4939-a953-6e72863befe9)
@laurenluz @divine-comedian 
https://www.figma.com/file/WCfheR8ID8yqNglhW1ZeM0/Project-Verification?type=design&node-id=6120%3A9383&mode=design&t=lQTSqn1b4Nte5saw-1

divine-comedian:
@Tosinolawale 

Just those 4 fields - you can remove things like website, youtube, instagram

We're looking for user handles not URLs so the default text in the input field should reflect that.

Tosinolawale:
![Project contact](https://github.com/Giveth/giveth-dapps-v2/assets/35389241/decbb498-6103-4d9e-86be-f2600813915a)
https://www.figma.com/file/WCfheR8ID8yqNglhW1ZeM0/Project-Verification?type=design&node-id=6120%3A9383&mode=design&t=E2QzqsiBBOhChcZP-1
@divine-comedian

divine-comedian:
@jainkrati  - Could you assign an FE dev when time permits for this one? This is adding the FE component to the project verification form - it will need some BE changes in order for the info from the form to be saved to the back-end https://github.com/Giveth/impact-graph/issues/1436

WhyldWanderer:
[...]

PR #4103, opened 2024-05-01
fix: project contact form modifs 

Meriem-BM:
Related to #3910 

- remove other project button
- add header Project Social link


MohammadPCh (approved):
@Meriem-BM thanks, lgtm

Issue #3675, opened 2024-01-24
Make it possible to edit QF mid-page banner in admin bro


laurenluz:
Every qf round, I make an issue for a dev to change this text. would be great to do it in admin bro directly.

from[this page](https://giveth.io/qf/all)

<img width="1388" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBL2IzQWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--be63b1b8edf53f5eac65b28be1e919d1eed258f8/image.png" alt="image.png" />

idk if this is a frontend or backend task @RamRamez @CarlosQ96 @jainkrati

RamRamez:
From front-end side it's a very easy task, I just need to read the title and description from an endpoint. But from the backend side, maybe backend guys can send title and description with the "allProjects" endpoint. @mohammadranjbarz @CarlosQ96 Is it possible?

mohammadranjbarz:
@CarlosQ96  Can you do it?

divine-comedian:
@mohammadranjbarz as per discussion take this one up

mohammadranjbarz:
@RamRamez Can you take this?
(If you had any question feel free to ask me)

CarlosQ96:
Leaving Pr in a few mins

CarlosQ96:
> From front-end side it's a very easy task, I just need to read the title and description from an endpoint. But from the backend side, maybe backend guys can send title and description with the "allProjects" endpoint. @mohammadranjbarz @CarlosQ96 Is it possible?

ok So no formatting, just the title and description, thats easy enough

laurenluz:
Yeah!

jainkrati:
@maryjaf pls verify and close

maryjaf:
I haven't understood which field in qf page be used for setting mid page banner ?
by changing "Description" in qf page from admin, nothing happens on QF banner
Are frontend changes needed?
@CarlosQ96 
<img width="1525" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMU9GQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--4d3e971d6e2ce9938407829c6edc527ae326ae1f/image.png" alt="image.png" />

[...]

PR #4108, opened 2024-05-02
fix: retrieve title and description for active QF Round banner

Meriem-BM:
Related to #3675


MohammadPCh (approved):
@Meriem-BM thanks, LGTM

PR #4133, opened 2024-05-08
Remove users field from project if it's not needed

RamRamez:
https://github.com/Giveth/impact-graph/issues/583


alireza-sharifpour (approved):
Tnx @RamRamez

PR #4134, opened 2024-05-09
first fixes w/ errors

divine-comedian:
There are some img tags inside of components we don't use anymore, such as gitcoin grants, referral program, and fiat purchase 

Also some other smaller elements that I couldn't find being used anywhere in the dapp. I left these unchanged since they are not used.


MohammadPCh (approved):
@divine-comedian Thanks, LGTM. I didn't check the images one by one on the website, but the code is fine.

PR #4139, opened 2024-05-10
Network column inside user donation

### LINKED ISSUE #4138: Add network column to donations table on user page
Currently while looking at the private and public view of the user's donation table there is no column that identifies on which network a donation was made:

Screenshot from 2024-05-09 18-39-36

We should add a column to this table and show the network logo for each donation to allow the user to know which network they have made their donations on. Similar to the recurring donations table: 

image

kkatusic:
Resolve #4138

@RamRamez, cc @jainkrati, when you have time please can you review this pull request.

Only problem that I have now is how to check _private view_ is it working like it should, to check table view with _status_ column included. Is there any option to log on development side with some other test user and check admin/private options?

Thx

profile


RamRamez (approved):
Great! Thanks @kkatusic

MohammadPCh (approved):
@kkatusic Thanks, LGTM

kkatusic:
> @kkatusic Thanks, LGTM

Thx @MohammadPCh for all help ;)

PR #4182, opened 2024-05-20
fix giveth tag for subgraph queries

RamRamez:
I used native fetch function instead of our custom gqlRequest function, since it's not supporting params for graphql request

https://github.com/Giveth/impact-graph/issues/934


kkatusic (commented):
Tested this branch, staked 10 GIV and lock 2.5 GIV for 6 month, 4x power and result is 17,5. It is ok on my side

kkatusic (approved):
Approving after review

PR #4184, opened 2024-05-20
add qf redirects only on production

MohammadPCh:


Meriem-BM (approved):
Thanks @MohammadPCh, LGTM!

PR #4229, opened 2024-06-02
Bypass maintenance mode

### LINKED ISSUE #4113: Add back functionality to bypass maintenance mode splash page
When the maintenance mode splash page is up we should have a way to bypass it to continue debugging on production. 

Previously we allowed it by clicking in the corner of the page to bypass the splash page and use the website normally. 

image

kkatusic:
Resolve #4113 

@divine-comedian, cc @jainkrati, added here solution for bypassing maintenance mode. It works using local storage variable _bypassMaintenance: true_ if you don't have that variable set up you can skip maintenance mode by clicking on top left corner on transparent button.

@divine-comedian , I checked old commits and old repository and didn't find that there was some bypassing enabled, but only checked _app.tsx and old index.tsx component. 

## Summary by CodeRabbit

- **New Features**
  - Introduced maintenance mode functionality, allowing users to be notified when the app is under maintenance.
  - Added a bypass option for maintenance mode, enabling users to continue using the app if they have special permissions.


divine-comedian:
@kkatusic - can you explain then what would a non-developer have to do in order to access the website when the maintenance mode is up?

kkatusic:
> @kkatusic - can you explain then what would a non-developer have to do in order to access the website when the maintenance mode is up?

@divine-comedian non-developer would click on the transparent button in the top left corner of maintenance page, like you did before

divine-comedian:
@kkatusic Awesome! thanks for that

kkatusic:
> @kkatusic Awesome! thanks for that

Np, thx you @divine-comedian for all support, I hope you tested ;)

MohammadPCh (approved):
@kkatusic Thanks, LGTM

PR #4231, opened 2024-06-02
fix: (case-sensitive issue) rename existing icon files

Meriem-BM:
- AGEUR -> agEUR
- RETH -> rETH


RamRamez (approved):
Thanks @Meriem-BM

Issue #4223, opened 2024-05-31
Support Giveth Subgraphs as ENV Variables


mhmdksh:
We need to have the subgraphs used in the Frontend as ENV Variables.

This is a requirement to move to subgraph studios, since we'll be accessing the subgraphs with API Keys in the URL

The subgraphs are currently hardcoded here: https://github.com/Giveth/giveth-dapps-v2/tree/develop/src/config

Also we need to be able to change them quickly to be able to test new subgraphs

kkatusic:
@mhmdksh there aren't one subgraph address inside configuration file, we need to remove one or two specific or all? thx

mhmdksh:
Hey @kkatusic we mainly need the below subgraphs to be identified in the ENV Variables:
```
line: 140,141
		subgraphAddress:
			'https://api.studio.thegraph.com/query/40764/giveconomy-staging-goerli/1.5.0?source=giveth',
```
```
Line: 196,197
		subgraphAddress:
			'https://api.studio.thegraph.com/query/40764/giveconomy-staging-gnosischain/1.5.1?source=giveth',
```
```
Line: 358, 359
		subgraphAddress:
			'https://api.thegraph.com/subgraphs/name/giveth/giveth-economy-optim-sepolia?source=giveth',
```

RamRamez:
@mhmdksh four new env var are added:

NEXT_PUBLIC_SUBGRAPH_MAINNET
NEXT_PUBLIC_SUBGRAPH_GNOSIS
NEXT_PUBLIC_SUBGRAPH_OPTIMISM
NEXT_PUBLIC_SUBGRAPH_CLASSIC

kkatusic:
> Hey @kkatusic we mainly need the below subgraphs to be identified in the ENV Variables:
> 
> ```
> line: 140,141
> 		subgraphAddress:
> 			'https://api.studio.thegraph.com/query/40764/giveconomy-staging-goerli/1.5.0?source=giveth',
> ```
> 
> ```
> Line: 196,197
> 		subgraphAddress:
> 			'https://api.studio.thegraph.com/query/40764/giveconomy-staging-gnosischain/1.5.1?source=giveth',
> ```
> 
> ```
> Line: 358, 359
> 		subgraphAddress:
> 			'https://api.thegraph.com/subgraphs/name/giveth/giveth-economy-optim-sepolia?source=giveth',
> ```

Thx @mhmdksh , @RamRamez made pull request

[...]

PR #4233, opened 2024-06-03
Support Giveth Subgraphs as ENV Variables

RamRamez:
#4223

## Summary by CodeRabbit

- **Chores**
  - Updated subgraph addresses to use environment variables for enhanced configurability and flexibility across different chains (mainnet, xDai, Optimism, Ethereum Classic).

### DIFF
--- a/src/config/development.tsx
+++ b/src/config/development.tsx
@@ -89,7 +89,7 @@ const classic = {
 			url: 'https://etc-mordor.blockscout.com',
 		},
 	},
-	subgraphAddress: 'http://167.172.97.150:8000/subgraphs/name/giveth/etc',
+	subgraphAddress: process.env.NEXT_PUBLIC_SUBGRAPH_CLASSIC,
 };
 
 const EVM_CHAINS = [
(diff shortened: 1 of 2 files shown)


mhmdksh:
@RamRamez Can you please follow the suggestions by coderabbit.

We should put a default value, if we removed the values that are in place now. Since if there was no default value (Or for any reason we didn't configure it, as is the case in the current setup), the frontend might not start.

kkatusic (approved):
Yes, LGTM ;)

mhmdksh:
> Yes, LGTM ;)

@kkatusic If this is merged as it is now, the frontend might not start with the current config

RamRamez:
> @RamRamez Can you please follow the suggestions by coderabbit.
> 
> We should put a default value, if we removed the values that are in place now. Since if there was no default value (Or for any reason we didn't configure it, as is the case in the current setup), the frontend might not start.

Done!

mhmdksh (approved):
These look good now, Thanks @RamRamez great work :)

RamRamez:
> These look good now, Thanks @RamRamez great work :)

But coderabbit is not convinced yet :))

Issue #3829, opened 2024-03-06
in the UI, we should explain to people that donations less than $1 in QF don't count


laurenluz:
@mosaeedi - can you help me out w/ this one?

mosaeedi:
<img width="564" alt="Screenshot 2024-04-22 at 5 17 25 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/13708322/fea3c7b0-09b4-4ced-9401-2f0e213973dd">


When the user enters an amount under $1 a hint can appear to let the user know.

mosaeedi:
@laurenluz wdyt?

laurenluz:
@mosaeedi I think this is good - we need to consider a couple of cases though...

- the project must be in the QF round
- we might want to hide the "estimated matching" amount if the amount is less than $1.

for ease of development, can you make a couple of versions here so the dev can understand the scenarios?

like:
1. Donation to QF project if the USD amount is >= threshold (FYI we set a min donation amount in our backend so this value should be pulled from there)
2. Donation to QF project if the USD amount is < threshold

mosaeedi:
Hey @laurenluz

I don't think how I can show the scenarios on Figma other than how it is right now!

The _yellow notif row_ "**Make it count**" will only appear if both scenarios are valid!

The project should be in **QF** and the **Donation should be less than the threshold**, 

and by default, **when the amount is less than the threshold**, we only show "**Make it count**" yellow notif row instead of the "estimated matching" green notif row.

Is it clear?

maryjaf:
- [ ] When the round hasn't started yet, should this explanation about minimum amount of donation be shown ?
 
 @mosaeedi @laurenluz 
 
![image](https://github.com/Giveth/giveth-dapps-v2/assets/111529185/c0d6854c-7531-47d1-80a1-865a6cb2a46c)

https://github.com/Giveth/giveth-dapps-v2/assets/111529185/ebc8311b-f805-41e8-9dbf-b8619d81623f

mosaeedi:
The "**Make it count**" explanation is only for the donation of less than a dollar **during the QF round**.

Meriem-BM:
> * [ ]  When the round hasn't started yet, should this explanation about minimum amount of donation be shown ?
> 
> @mosaeedi @laurenluz
> 
[...]

PR #4297, opened 2024-06-19
fix: keep minimumValidUsdValue field only on FETCH_PROJECT_BY_SLUG_DO…

Meriem-BM:
…NATION query

Related to #3829

## Summary by CodeRabbit

- **Refactor**
  - Removed the `minimumValidUsdValue` field from project data queries to simplify the data structure.

- **Bug Fixes**
  - Made the `minimumValidUsdValue` property optional in the `IQFRound` interface for better flexibility and to handle potential null values.


RamRamez (approved):
Thanks @Meriem-BM

PR #4332, opened 2024-06-25
Estimated matching range

Meriem-BM:
- Format price on format donation for rounded case set to true
- add `getEstimatedMatchingRange` to global QF helpers
- format price for Archived QF table column Matching Pool


RamRamez (approved):
Thanks @Meriem-BM

Issue #4334, opened 2024-06-25
Copy fixes on QF


laurenluz:
- [ ] Donate opage, "donation is less than threshold" box - this should show the min threshold with 2 decimal places, i.e. $0.90
![image](https://github.com/Giveth/giveth-dapps-v2/assets/75490971/1e21c53f-f0df-4ab0-83c2-da3ba38d835c)

PR #4346, opened 2024-06-25
Fix: format min USD value for estimated matching to be counted

Meriem-BM:
Related to #4334

### DIFF
--- a/src/components/views/donate/EstimatedMatchingToast.tsx
+++ b/src/components/views/donate/EstimatedMatchingToast.tsx
@@ -88,13 +88,26 @@ const EstimatedMatchingToast: React.FC<IEstimatedMatchingToast> = ({
 			)} ${allocatedFundUSDPreferred ? '' : ` ${allocatedTokenSymbol}`}`
 		: '---';
 
+	const formatMinUSDValue = (
+		num: number | undefined,
+		precision: number = 2,
+	) => {
+		if (!num) return '0';
+
+		if (num >= 1 || num < 0.01) {
+			return num.toString();
+		}
+
+		return num.toFixed(precision);
+	};
+
 	const bottomText = isAboveMinValidUsdValue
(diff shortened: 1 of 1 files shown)


src/components/views/donate/EstimatedMatchingToast.tsx
```diff
@@ -88,13 +88,26 @@ const EstimatedMatchingToast: React.FC<IEstimatedMatchingToast> = ({
...
+	};
+
 	const bottomText = isAboveMinValidUsdValue
 		? formatMessage({ id: 'page.donate.matching_toast.bottom_valid' })
 		: formatMessage({
 				id: 'page.donate.matching_toast.bottom_invalid_p1',
 			}) +
 			' $' +
-			minimumValidUsdValue +
+			formatMinUSDValue(minimumValidUsdValue) +
```
RamRamez, line 110: I suggest using `formatBalance(minimumValidUsdValue)` instead

RamRamez (changes requested):

src/components/views/donate/EstimatedMatchingToast.tsx
Meriem-BM: @RamRamez I made that change, but I think `minimumValidUsdValue` can be also more less than `0.000001` so we will see `0.00` on FE, I think we should either add some min input validation on BE or WDYT?

kkatusic (approved):
LGTM ;)

src/components/views/donate/EstimatedMatchingToast.tsx
RamRamez: @Meriem-BM It's USD value not token amount. I think setting `minimumValidUsdValue` less than 0.000001 USD doesn't make sense because it's almost zero.

RamRamez (approved):
Thanks @Meriem-BM

Issue #4204, opened 2024-05-23
"Last Update" on project card is not correct


WhyldWanderer:
This project made an update today but it still says "Last update 22 days ago" on the project card.
https://giveth.io/project/digital-street-musician-doing-free-live-concerts

![image](https://github.com/Giveth/giveth-dapps-v2/assets/74267818/1254a83e-f9c5-4a47-8636-3af752e7a326)

![image](https://github.com/Giveth/giveth-dapps-v2/assets/74267818/60baf5f5-7eaf-41cf-960f-afac50a141d2)

divine-comedian:
@CarlosQ96 

I checked several projects after Ashley's report and it appears this number is not updating at all for any project.. who knows since when.

divine-comedian:
@CarlosQ96  Can you please check and provide an update on this issue.

RamRamez:
@divine-comedian You need last update field on project card to be updated after only on adding a project update? So what about editing description and title?

WhyldWanderer:
@RamRamez - jumping in to answer the question. ..

It should only show the time that the last update was added to the project.

You can ignore changes to project description/title

RamRamez:
Thanks @WhyldWanderer 
I can take this issue.

RamRamez:
@maryjaf @LatifatAbdullahi Please test and verify

maryjaf:
I've tested and now the "Last Updated" on project card and "Latest Updates" in Homepage are shown correctly.

PR #4402, opened 2024-07-10
"Last Update" on project card is not correct

RamRamez:
#4204

## Summary by CodeRabbit

- **New Features**
	- Updated project cards to display the latest update creation date instead of the last updated time.

### DIFF
--- a/src/apollo/gql/gqlProjects.ts
+++ b/src/apollo/gql/gqlProjects.ts
@@ -34,6 +34,7 @@ export const PROJECT_CARD_FIELDS = gql`
 			avatar
 		}
 		updatedAt
+		latestUpdateCreationDate
 		organization {
 			label
 		}
--- a/src/apollo/types/types.ts
+++ b/src/apollo/types/types.ts
@@ -63,7 +63,7 @@ export interface IProject {
 		name?: EProjectStatus;
 	};
 	updatedAt: string;
-	latestUpdateCreationDate: string;
+	latestUpdateCreationDate?: string;
 	organization?: {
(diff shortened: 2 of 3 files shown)


src/apollo/types/types.ts
```diff
@@ -63,6 +63,7 @@ export interface IProject {
 		name?: EProjectStatus;
 	};
 	updatedAt: string;
+	latestUpdateCreationDate: string;
```
ae2079, line 66: Is it not better to make this field optional? (I think this help safer deploy and prevent crashes)

src/apollo/gql/gqlProjects.ts
```diff
@@ -34,6 +34,7 @@ export const PROJECT_CARD_FIELDS = gql`
 			avatar
 		}
 		updatedAt
+		latestUpdateCreationDate
```
ae2079, line 37: Does the backend serve this field now?

ae2079 (commented):
Thanks @RamRamez

ae2079 (approved):
Looks like the preview link works correctly

src/apollo/gql/gqlProjects.ts
RamRamez: Yes

src/apollo/types/types.ts
RamRamez: I have added `latestUpdateCreationDate` to project table, just like updatedAt above that is required field.

RamRamez: I made that optional

Issue #3420, opened 2023-11-21
Error on Staking with Permit xDAI/FOX farm -  NOT_VALID_CALL_SIGNATURE


divine-comedian:
**Describe the bug**
When users try to stake an LP in the xDAI/FOX farm the transaction is failing - The error is coming from the UnipoolTokenDistributor contract and says `UnipoolTokenDistributor: NOT_VALID_CALL_SIGNATURE`

 https://dashboard.tenderly.co/tx/gnosis-chain/0xb61995a18f956bf81c7e535c29d30463339bb667247f35aafeaaf0fe05336bd1

**To Reproduce**
1. Go to GIVfarm on xDAI
2. Create xdai/fox lp on honeyswap
3. Click stake 
4. toggle 'Permit Mode' on
5. stake some tokens
6. see error in metamask
7. proceed anyway
8. see failed tx in tenderly

**Expected behavior**
The user should be able to successfuly stake their LP tokens with Permit mode enabled

**Screenshots**
![image](https://github.com/Giveth/giveth-dapps-v2/assets/67759413/72a53536-1da9-4506-b03e-3d1c9e7c6481)

**Desktop (please complete the following information):**
 - OS: Ubunutu 22.04
 - Browser: Brave
 - Wallet: MetaMask

**Additional context**
Reported on production by Shapeshift users

divine-comedian:
cc: @laurenluz

laurenluz:
the rewards on the fox/xDAI farm I believe have ended as of yesterday from [this PR](https://github.com/Giveth/giv-token-contracts/pull/90) they started on May 30.

would have been great to see this issue before! what's the story here?

laurenluz:
I mean that I see Krati was assigned, then Ramin, then after my comment Ramin set an estimate... so it looks to me like there is some conversation & coordination going out outside of this issue that I'm not in the loop on... 

This farm needs to be archived. I will make a new issue for that... but we should disable the stake function anyway, so this issue is kind of irrelvant unless we deploy a new regen farm or reactive/extend this.

divine-comedian:
> I mean that I see Krati was assigned, then Ramin, then after my comment Ramin set an estimate... so it looks to me like there is some conversation & coordination going out outside of this issue that I'm not in the loop on...
> 
> This farm needs to be archived. I will make a new issue for that... but we should disable the stake function anyway, so this issue is kind of irrelvant unless we deploy a new regen farm or reactive/extend this.

[...]

PR #4405, opened 2024-07-11
Fix two sign request on LP staking

kkatusic:
Resolving #3420 

Please @RamRamez can you double check this fix, first time I'm doing something about hash and staking function don't want to mess it up. If you need some tokens to test, be free to dm me 

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of permit signatures during token staking to ensure a smoother and more reliable user experience.


RamRamez (approved):
Thanks @kkatusic

Issue #4247, opened 2024-06-05
User flow for implementing model-based detection system into QF rounds


laurenluz:
We are adding a 2-part user verification model for qf eligibility. Backend is explained in https://github.com/Giveth/giveth-dapps-v2/issues/4221

This issue is for design. The user flow is essentially:
- At the start of a new QF round, every donor must verify/re-verify their eligibility
- The first "check" is called either by a button click "check your eligibility" or by the user making their first donation. In the backend, we check if their address passes... if they do, the user sees "you're eligible!" and no other info about passport/stamps, etc.
- If the user fails the 1st check (with the model based system), they are passed through our gitcoin passport flow, where they can verify their stamps

We will have to consider:
- banners on QF page
- donation success page eligibility box
- user profile
- giveth.io/passport page (we may want to revamp this into a qf eligibility page)

@mosaeedi @koday1 cc @CarlosQ96

mosaeedi:
## Here is the first design draft

Please take a look at the proposed structure and designs for the Profile page.


# Structure 

So with the new method in mind, I create two sections for the profile page.

**1. User Verification**
On the first section, User can verify with the MBD method, or [in case they couldn't pass the MBD] use Gitcoin Passport.


**2. QF Donation Eligibility**
This section is more like an informative section just to make it crystal clear for user either they are eligible or not.


# Designs 
Now let's dive into the designs!


## User A
Users A, are those who don't have a connected Gitcoin Passport and are Not verified using the MBD method.
For these users, they can only see the MBD card and the Gitcoin passport card is grayed out and disabled.

<img width="811" alt="Screenshot 2024-06-11 at 4 55 04 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/13708322/e327d5ad-7c5f-4797-9f40-c773078a8ee5">

### User A-1 [MBD Verified]

When they pass the MBD check successfully, the Gitcoin Passport is get enabled, but on the QF eligibility card, they can see that, they are already eligible for QF matching.

### User A-2 [MBD Failed]

In case the user fails the MBD check, the Gitcoin passport is get enabled as an alternative way for user to verify.

<img width="815" alt="Screenshot 2024-06-11 at 5 02 10 PM" src="https://github.com/Giveth/giveth-dapps-v2/assets/13708322/abd8b2aa-f087-4d27-843d-ece274e4ca67">

### User A-3 [MBD Failed, Passport Checked]
And when they passed the Gitcoin check...

[...]

PR #4501, opened 2024-07-30
Fix: show error banner after checking round

Meriem-BM:
Related to #4247


MohammadPCh (approved):
Thanks @Meriem-BM LGTM

Issue #3903, opened 2024-03-27
Deduct transaction fee if user tries to donate/deposit max token balance


divine-comedian:
To prevent users from accidentally donating or depositing all of their native tokens we should automatically subtract the estimated transaction fee from their available balances
 
If a user presses the button to input the max amount into an input field such as depositing into their stream balance or donating to a project we should input their native balance - estimated tx fee

If a user manually inputs an amount that exceeds this projected available balance we should throw a warning underneath the input field and prevent the user from continuing - they must lower the amount to continue 

The warning text should say: 

"You need to keep at least {amount} {native token} in your wallet to pay the network fees for this transaction. Please enter a lower amount."

---
to be clear this is a very niche error message, take for example: 

- user balance is 10 tokens
- they choose to donate 9.9
- network fee is 0.2
- the total cost of the tx would be 10.1

 which exceeds their balance, even though they are "donating" less than their balance the tx fee added would cause it to fail. so we show the error message.

----
 
 > @divine-comedian @maryjaf All issues have been resolved except case 11. @maryjaf can you reproduce it again? 
 > let's have a call and reproduce it together. but you can check all other issues on the staging environment.
> And case 11 happens when for example  the user has no super token and in donation flow after getting super token the > user's ETH balance change to lower than minimum for a donation transaction 
> in this scenario the above message is shown that it would be better a handle message be shown

_Originally posted by @maryjaf in https://github.com/Giveth/giveth-dapps-v2/issues/3091#issuecomment-2019840839_

divine-comedian:
Are we still waiting to pick this one up?

RamRamez:
@Reshzera could you please take it?

Reshzera:
@RamRamez Yes, I'll take a look

Reshzera:
@maryjaf @LatifatAbdullahi This Issue is ready to be tested on the staging, could you please take a look?

maryjaf:
Could you please take a look on below screen record ? @Reshzera 

- by entering the number in box, the available amount is changed 

https://github.com/user-attachments/assets/470f64d7-eba0-4999-ae91-bf66e58f77e4

Reshzera:
@maryjaf Sure

Reshzera:
@maryjaf I just push a new PR to staging to correct this bug, could you please take a look?

maryjaf:
> Could you please take a look on below screen record ? @Reshzera
> 
> * by entering the number in box, the available amount is changed

[...]

PR #4528, opened 2024-08-01
Feat/issue 3903

Reshzera:
## Summary by CodeRabbit

- **New Features**
	- Improved error handling for donations by focusing on insufficient gas fees instead of generic amount errors.
	- Enhanced UI updates to reflect accurate balance information related to gas fees during the donation process.

- **Bug Fixes**
	- Addressed validation issues to ensure donations are accurately processed based on the user's balance and gas fees.

### DIFF
--- a/src/components/views/donate/OnTime/OneTimeDonationCard.tsx
+++ b/src/components/views/donate/OnTime/OneTimeDonationCard.tsx
@@ -88,7 +88,8 @@ const CryptoDonation: FC = () => {
 	const [amount, setAmount] = useState(0n);
 	const [erc20List, setErc20List] = useState<IProjectAcceptedToken[]>();
 	const [anonymous, setAnonymous] = useState<boolean>(false);
-	const [amountError, setAmountError] = useState<boolean>(false);
+	const [insufficientGasFee, setInsufficientGasFee] =
+		useState<boolean>(false);
 	const [showDonateModal, setShowDonateModal] = useState(false);
 	const [showInsufficientModal, setShowInsufficientModal] = useState(false);
 	const [showChangeNetworkModal, setShowChangeNetworkModal] = useState(false);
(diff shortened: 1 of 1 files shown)


src/components/views/donate/OnTime/OneTimeDonationCard.tsx
```diff
@@ -311,18 +312,21 @@ const CryptoDonation: FC = () => {
 
 	useEffect(() => {
 		if (
-			amount + gasfee > selectedTokenBalance &&
+			amount > selectedTokenBalance - gasfee &&
+			amount < selectedTokenBalance &&
```
lovelgeorge99, line 316: i dont think you need this condition `amount  < selectedTokenBalance` because it is going to make the condition to false

Reshzera: # Explanation of Transaction Conditions

The removal of the original condition and the addition of the two new conditions are necessary to cover specific scenarios of balance and gas fee more precisely. The condition `amount < selectedTokenBalance` is needed to ensure the transaction amount does not exceed the total balance, while `amount > selectedTokenBalance - gasfee` ensures the transaction is not approved if, after subtracting the gas fee, the remaining balance is not enough to cover the transaction. This prevents the transaction from failing due to insufficient balance to cover the gas fee.

Therefore, both additional conditions are necessary to ensure the transaction is only approved when there is enough balance to cover both the amount and the gas fee, avoiding potential insufficient balance issues. Also, if the transaction exceeds the wallet's balance, a modal will be displayed to inform the user about the insufficient balance. Because this modal will be shown, there is no need to show this error in another way.

### Scenario Explanation

Assume:
- `selectedTokenBalance` = 100 tokens
- `gasfee` = 5 tokens
- `amount` = 95 tokens

- If `amount` were 96 tokens:
  - First condition: 96 tokens > 95 tokens (true)
  - Second condition: 96 tokens < 100 tokens (true)
  - The warning will be shown.

- If `amount` were 105 tokens:
  - First condition: 105 tokens > 95 tokens (true)
  - Second condition: 105 tokens < 100 tokens (false)
  - The modal of insufficient balance will be shown.

### Conclusion

Both conditions ensure:
1. The `amount` does not exceed the total balance.
2. The remaining balance after the gas fee is enough to cover the transaction.

If either condition fails, a modal informs the user about the insufficient balance. Therefore, there's no need to display the error in another way.

lovelgeorge99 (approved):

Reshzera:
#3903

Issue #4577, opened 2024-08-15
Make Recurring Donations default option


divine-comedian:
We would like to further promote the experimentation by users of the recurring donation feature. With that we should show the recurring donations view by default when the user loads the donate page IF the project has added an optimism address.

![Image](https://github.com/user-attachments/assets/d01284dc-a0d2-4cbc-9e10-2b57efa4b87e)

## Requirements
- when entering donate page for a specific project WITH an Optimism recipient address the recurring donation view is shown by default on page load
- entering the donate page for a project WITHOUT an Optimism recipient address shows the one-time donation view.

divine-comedian:
@kkatusic think you can take this one on?

maryjaf:
> * when entering donate page for a specific project WITH an Optimism recipient address the recurring donation view is shown by default on page load
> * entering the donate page for a project WITHOUT an Optimism recipient address shows the one-time donation view.

Thanks @kkatusic These ACs are passed

PR #4623, opened 2024-08-26
Feature: added recurring tab as default for OP address

kkatusic:
- https://github.com/Giveth/giveth-dapps-v2/issues/4577


divine-comedian:
@MohammadPCh @mateodaza @RamRamez - if we can review this one with a higher priority that would be great! I think this small change will make a big difference in getting people to try out this feature.

Meriem-BM (approved):
Thanks @kkatusic, LGTM!

Issue #4557, opened 2024-08-13
refreshed the donation page and got an application error


GriffGreen:
<img width="1553" alt="Screenshot 2024-08-12 at 10 01 17 PM" src="https://github.com/user-attachments/assets/90059156-b0aa-4dfa-9a90-f31dd39f76ce">

jainkrati:
@maryjaf @RamRamez could you please take a look at this issue : how frequently it happens ? what is the cause ? deploy the fix sooner

HrithikSampson:
I was able to reproduce this issue using URL Throttler extension and setting a delay to https://rpc.walletconnect.org/* for a long time while refreshing the page and I clicked on select Token very quickly I got this error in giveth.io

<img width="1447" alt="Screenshot 2024-08-15 at 3 20 42 PM" src="https://github.com/user-attachments/assets/8a193ac9-e62c-4882-b5ca-2b88f6b89181">

Similarly In localhost I got the following error

<img width="1075" alt="Screenshot 2024-08-15 at 6 31 57 PM" src="https://github.com/user-attachments/assets/7b780bc6-f0cf-4234-9876-3a2aa4b6e187">

divine-comedian:
@HrithikSampson - What is the update on this issue?

HrithikSampson:
Hi @divine-comedian ,

I was facing errors while trying to solve it. There was one solution where I was able to get resolve the error but it was causing many rerenders to  to the page.

I will try another solution to it. I am not sure whether it will work.

HrithikSampson:
Hi @RamRamez @divine-comedian ,

I was getting a react queueing issue previously(it wasnt coming all the time) but after a small change the queueing issue disappeared. I cant see it now but I am not sure why it isnt coming. I tried it multiple times and it isnt coming.

I was getting this error:
Should have a queue. This is likely a bug in React. Please file an issue

kkatusic:
@HrithikSampson, cc @divine-comedian, maybe this is caused ny same issue that Ashley reported, searching inside input field for token. 
If that is true @lovelgeorge99 is handling this.

HrithikSampson:
@kkatusic ,

I think root cause might be the same.

I just wanted to inform that I am not seeing any error coming while clicking select Token Modal while refreshing now. But I am getting the error which Ashley reported while searching for the tokens.

@divine-comedian, I think that if the error is resolved while searching for the tokens, this error will also be resolved along with it.

[...]

PR #4633, opened 2024-08-28
Feat/issue 4557

HrithikSampson:
relates to #4557 

Summary

1. Since in some of the cases the hooks were being redered more or less than the previous render. I have put the balances of all the tokens coming into SelectTokenInnerModal in a map and fetching it through the map.

2. React Queueing issue was occuring sometimes while using useFetchBalance directly but the error disappeared while using fetchBalance instead of useFetchBalance.

## Summary by CodeRabbit

- **New Features**
	- Improved handling of token balances in the Select Token Modal for enhanced performance and robustness.
	- Added support for Solana wallet integration, allowing for better token balance management.
	- Introduced a new function for retrieving token balances across different blockchain environments.

- **Bug Fixes**
	- Enhanced error handling for scenarios where token data may be undefined, preventing potential runtime errors.
	- Improved type safety in balance calculations to reduce the risk of type mismatches.


lovelgeorge99 (approved):
LGTM 👍

Issue #4671, opened 2024-09-03
app is crashing when user enters something in token search box on donation page


divine-comedian:
on PRODUCTION if you type anything into the token search box the application throws an error and crashes. 

![Image](https://github.com/user-attachments/assets/39797609-7262-47c7-9da8-55e52d47ee25)

![Image](https://github.com/user-attachments/assets/28ac9949-5b6c-4b53-9553-d3035ab57d7a)

PR #4672, opened 2024-09-03
removing need to call hooks in a loop instead call getBalanceForToken

HrithikSampson:
I had stashed my commits after a soft reset and then popped it on this branch which was branched from main branch.

So all of the changes are coming in a single commit.

- https://github.com/Giveth/giveth-dapps-v2/issues/4671


kkatusic (approved):
thx @HrithikSampson

PR #4710, opened 2024-09-09
hotifx: prevent projects with Endaoment label from showing recurring donation view

divine-comedian:
modify condition from https://github.com/Giveth/giveth-dapps-v2/pull/4623/files to check if projects do not have endaoment label in order to show recurring donation view in donation flow


mateodaza (approved):
LGTM

PR #4823, opened 2024-10-04
hotfix: text update for non-verified no status projects

divine-comedian:


RamRamez (approved):
Thanks @divine-comedian

PR #4831, opened 2024-10-08
Fix/usdt donation on mainnet

kkatusic:


MohammadPCh (approved):
Thanks @kkatusic LGTM

Issue #915, opened 2022-06-12
Create Sitemap for Better SEO


MohammadPCh:
We need to generate a sitemap for our website to search engines can find our pages better and easier. now they can only crawl pages linked to our popular pages (homepage or first load of projects).
I suggest creating the sitemap once per day. We need the devOps team's help with this.
@aminlatifi @MoeNick

MohammadPCh:
This google doc for Sitemap:
https://developers.google.com/search/docs/advanced/sitemaps/overview

divine-comedian:
@mhmdksh @MohammadPCh - is this issue still relevant? 

I'm VERY interested in improving our SEO.

divine-comedian:
@mhmdksh can you please provide an update on this issue?

mhmdksh:
> @mhmdksh can you please provide an update on this issue?

@divine-comedian Sorry I can't provide any update on this neither this is my field of expertise. I prefer if I get unassigned as this is a Frontend/SEO-Optimization issue, not Devops work.

Thanks

@MohammadPCh Anyone can take this??

divine-comedian:
@MohammadPCh can this issue be passed to someone or what is the next course of action?

MohammadPCh:
@kkatusic @lovelgeorge99 @Reshzera guys who can take it?
we need a script to get projects from the server, add the static pages, and create a sitemap.
let me know if you need help

kkatusic:
> @kkatusic @lovelgeorge99 @Reshzera Hey everyone, who has the bandwidth to take this on?
> 
> We need a script to:
> 
> 1. Retrieve projects, users, and campaigns from the server
> 2. Add the static pages
> 3. Create a sitemap
> 
> Let me know if you need any help.

I will take it, thx @MohammadPCh

kkatusic:
@divine-comedian and @MohammadPCh, cc @CarlosQ96  is it ok to expose users inside sitemap?

divine-comedian:
I am not sure about what the implications are so I would defer to what either of them think

kkatusic:
Let's than wait @MohammadPCh and @CarlosQ96 responsne ;)

MohammadPCh:
@kkatusic Yes we can add the public profile to the sitemap. but we should disallow the dashboard and private zones.

CarlosQ96:
Ok we should make a query just to expose the minimum required info of users.
[...]

PR #4847, opened 2024-10-22
Feat/sitemap seo

kkatusic:
- https://github.com/Giveth/giveth-dapps-v2/issues/915

Just to notice here, this should be run by corn job, one BE or FE it's same. It's using Bearer Token for running and calling API endpoint: `giveth.io/api/generate-sitemap `
Example wit _curl_ call is:

```
curl -X GET http://localhost:3000/api/generate-sitemap \
  -H "Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCIAAAA9.JleHAiOjE3MzA3MzAwMjZ9.PPh8Y42iwTVs4poCfdtsdgdsScxE8bw83JKHlRfqjcLfrO0q5AJ8"
```

also that token should be defined on both end inside _.env_ file with variable: `SITEMAP_CRON_SECRET`


divine-comedian:
@mateodaza @CarlosQ96 can either of you pick up and review this PR please

CarlosQ96 (approved):
We can start like this, since its once a week.... But @divine-comedian I would consider redoing this, so that backend provides the xml completely in one go, maybe upload the file somewhere for frontend to fetch. Too many request can potentially crash the app. 
Run it in non peak hours.

PR #4859, opened 2024-11-10
add native usdc fix text on bridged USDC

divine-comedian:
Bug reported by superfluid team 

Fix incorrect labelling of bridged usdc as native usdc 

add native USDC as eligible recurring donation token


kkatusic (approved):
@divine-comedian you changed token hash it is ok, it is last one hash

PR #4904, opened 2024-12-23
Adding last recurring donation branchs

kkatusic:


mateodaza (commented):
@kkatusic I can't see the vercel logs, but I would infer it's related to the branch pointing  to staging instead of production - could you make sure this is the case before merging? the vercel deployment is failing

kkatusic:
> @kkatusic I can't see the vercel logs, but I would infer it's related to the branch pointing to staging instead of production - could you make sure this is the case before merging? the vercel deployment is failing

I don't know is it because setup of that branch. We can always revert last changes?

PR #4983, opened 2025-01-28
HOTFIX: update docs links + update hardcoded docs links

divine-comedian:
fix broken links on production as result of docs migration to notion


kkatusic (approved):
LGTM ;), thx @divine-comedian

PR #5055, opened 2025-03-10
Hotfix removing amazon instance

kkatusic:


CarlosQ96 (approved):

geleeroyale (commented):
thanks Kechy - I  think this should be enough to mitigate the attack vector as outlined in the disclosure document

Issue #5122, opened 2025-06-04
Cause Donation Success Page


divine-comedian:

divine-comedian:
@maryjaf FYI I added some more requirements and updated AC for this issue

maryjaf:
it seems this part is removed from project success donation page on prod 
should it be shown in cause success donation page ? @divine-comedian 

<img width="642" height="280" alt="Image" src="https://github.com/user-attachments/assets/3d5f0a66-a883-4f5c-98cd-d85f225ef3c2" />

<img width="2750" height="1574" alt="Image" src="https://github.com/user-attachments/assets/854f937d-c56f-4b21-b1e2-a6667ec18dc3" />

maryjaf:
> * remove "zero-fees" text component 
Pass

> * update bottom button text and destination "SEE MORE CAUSES" and it should redirect to the "all causes" page
Pass

> * share links should be updated to reflect copy defined in [cause success share link text and general share text for social media #5100](https://github.com/Giveth/giveth-dapps-v2/issues/5100)
pass

> * feedback form functionality from project donation success page should be implemented here, however the typeform link should be changed to the following: https://giveth.typeform.com/causefeedback

it should be tracked on this bug issue: https://github.com/Giveth/giveth-dapps-v2/issues/5173

> * cause card on left side of page should contain relevant information about cause including title, author name, description, total raised by cause and number of projects in cause.

Pass

<img width="2840" height="1630" alt="Image" src="https://github.com/user-attachments/assets/53e40810-873c-4e8e-8729-897c76c40a3a" />

divine-comedian:
I think this is the id 

`e68DoSqk` 

also here is the embed snippet I received from typeform dashboard, not sure if it is helpful 

```html
<div data-tf-live="01K0873JETTT7F94A7YWTFY6EV"></div><script src="//embed.typeform.com/next/embed.js"></script>
```

PR #5179, opened 2025-07-16
Cause feedback form

maryjaf:
add relevant id for cause feedback form
https://github.com/Giveth/giveth-dapps-v2/issues/5122#issuecomment-3076072515

## Summary by CodeRabbit

* **New Features**
  * The donation form now adapts its display based on a new URL parameter, showing either an embedded Typeform or the existing widget depending on the parameter value.


kkatusic (commented):
@maryjaf why you're not using widget here if we have same code for project?

PR #5233, opened 2025-08-01
Tweak select project ux

divine-comedian:
in select project view on create cause:

- remove learn more button on project card 
- stop height change of card on hover 
- add link open in new tab when hovering title or description
- on right hand select project list: show author name instead of 1st main category


kkatusic (commented):
These changes looks good, thx @divine-comedian

PR #5362, opened 2025-09-25
hotfix: add component breaking down fees

divine-comedian:
based on stakeholder feedback from @GriffGreen and @laurenluz put in a component on donation page displaying related cause fees


kkatusic (approved):
LGTM ;) , thx @divine-comedian

PR #5514, opened 2026-04-22
FIxing NaN on USer boosts and project boost

kkatusic:
## Summary by CodeRabbit

* **Bug Fixes**
  * Consistent display for zero/unknown GIVpower values (shows 0 or -- where appropriate)
  * Safer total and per-row GIVpower calculations to avoid NaN and incorrect totals
  * More robust balance aggregation by normalizing addresses for correct grouping
  * Stricter zero-balance notice conditions to reduce false prompts and only show when appropriate


ae2079 (approved):
thanks @kkatusic

Issue #5574, opened 2026-06-08
Estimated Matching Box Not Visible for Logged-Out Users


maryjaf:
Description:
When a user is not signed in, the Estimated Matching box is not displayed.

Current Behavior:

Logged-out users do not see the Estimated Matching box.
The box appears only when the user is signed in.

Expected Behavior:

The Estimated Matching box should be visible to users regardless of authentication status , please correct me if i'm wrong @MoeNick 
While actions related to Giveback require the user to be signed in, the estimated matching information should still be displayed to logged-out users.

Steps to Reproduce:

Open the application while logged out.
Navigate to a page where the Estimated Matching box is expected to appear.
Observe that the Estimated Matching box is missing.
Sign in and revisit the same page.
Observe that the Estimated Matching box is displayed.

<img width="1484" height="790" alt="Image" src="https://github.com/user-attachments/assets/ea532487-d888-46e4-971c-60ad6ff98406" />
<img width="1486" height="810" alt="Image" src="https://github.com/user-attachments/assets/a5850235-0bec-46bc-911b-9aed99d8a118" />

**Acceptance Criteria:**

- [x]  The Estimated Matching box is displayed for both logged-in and logged-out users.
- [x]  Logged-out users can view estimated matching values without being prompted to sign in.
- [x]  Giveback-specific actions and functionality remain restricted to authenticated users.
- [x]  The Estimated Matching box displays the same calculation and content for logged-out users as it does for logged-in users (where applicable).

cc: @laurenluz @kkatusic

kkatusic:
@maryjaf on normal donation process you need also to have connected wallet to see those values, for Stellar you have also nice notification: _Sign into Giveth for a chance to win GIVbacks._ .

Let us wait @MoeNick what will say.

MoeNick:
As no sybil attack prevention is essential in Stellar we excluded it, I think there's no need to limit it for only logged in users, we don't have log in requirement for stellar donations.

maryjaf:
[...]

PR #5577, opened 2026-06-09
Fix/5574 stellar estimated matching

kkatusic:
## Issue

#5574

## Summary

This PR fixes the Stellar QR donation flow so logged-out users can see estimated matching information without needing an EVM/Solana wallet connection. Estimated matching and QF eligibility are now based on the selected Stellar round, project data, and donation amount, while the existing sign-in requirement for donation submission and Givebacks-specific behavior remains separate.

## Changes

- Enabled QF round smart selection for Stellar QR donations by using the Stellar network ID instead of requiring a connected wallet chain.
- Updated Stellar QF round fallback selection to validate against the Stellar network when no smart-select round is returned.
- Removed the QR donation wallet-connection gate that prevented estimated matching from displaying for logged-out users.
- Show eligibility badges in the Stellar QR flow without requiring a connected wallet, matching the behavior needed for QR donations.
- Kept the estimated matching calculation tied to the selected Stellar QF round, token price, and donation amount.

## How to Test

Please verify and adjust these steps if needed:

1. Open a project that accepts Stellar QR donations and is eligible for an active QF round with Stellar support.
2. Log out or use an unauthenticated browser session.
3. Start a Stellar QR donation flow from the project donate page.
4. Enter an amount that meets the selected QF round minimum.
5. Confirm the eligibility badges and Estimated Matching box are visible without connecting a wallet or signing in.
6. Confirm the estimated matching value appears based on the entered amount.
7. Continue to the next step and verify sign-in/Givebacks behavior is still handled separately where required.

### DIFF
--- a/src/components/views/donate/DonationCardQFRounds/DonationCardQFRounds.tsx
+++ b/src/components/views/donate/DonationCardQFRounds/DonationCardQFRounds.tsx
@@ -86,11 +86,15 @@ export const DonationCardQFRounds = ({
 	const [showQFRoundModal, setShowQFRoundModal] = useState(false);
 
 	// Fetch QF round smart selection data
+	// Stellar (QR) donations are not tied to a connected EVM chain, so enable
+	// the query using the Stellar network number rather than requiring chainId.
 	const { data: smartSelectData, isFetching: isFetchingSmartSelect } =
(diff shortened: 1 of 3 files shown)


RamRamez (approved):
Thanks @kkatusic

PR #5578, opened 2026-06-10
Remove badges for non connected user on Stellar donation page

### LINKED ISSUE #5574: Estimated Matching Box Not Visible for Logged-Out Users

kkatusic:
## Description

Closes https://github.com/Giveth/giveth-dapps-v2/issues/5574

This PR updates the Stellar QR donation flow so logged-out donors can still see estimated matching information when their donation qualifies for an active QF round.

## Changes

- Show estimated matching for Stellar QR donations based on project, round, token price, and amount rather than wallet/sign-in state.
- Adjust Stellar donation eligibility badges so the logged-out QR flow can still surface QF eligibility where appropriate.
- Keep GIVbacks-specific messaging gated behind actual GIVbacks eligibility for both the project and selected Stellar token.
- Avoid prompting Stellar QR donors to connect a wallet for QF matching, since matching does not require an EVM/Solana wallet connection.

## Testing

- Not run locally.

## Summary by CodeRabbit

* **New Features**
  * Added a wallet sign-in prompt for GIVbacks opportunities during QR-code donations when both the project and selected token are eligible and the user can sign in.

* **Improvements**
  * Adjusted eligibility badge display for Stellar token donations so badges are suppressed when the GIVbacks sign-in prompt is shown, based on wallet connection and user status.

### DIFF
--- a/src/components/views/donate/OneTime/SelectTokenModal/QRCodeDonation/QRDonationCard.tsx
+++ b/src/components/views/donate/OneTime/SelectTokenModal/QRCodeDonation/QRDonationCard.tsx
@@ -8,6 +8,7 @@ import {
 	neutralColors,
 	IconArrowLeft,
 	mediaQueries,
+	IconWalletOutline24,
 	OutlineButton,
 	IconArrowRight16,
 	Button,
@@ -20,6 +21,7 @@ import { ethers } from 'ethers';
 import {
 	InputWrapper,
 	SelectTokenWrapper,
+	BadgesBase,
 	ForEstimatedMatchingAnimation,
 } from '../../../common/common.styled';
 import { TokenIconWithGIVBack } from '../../../TokenIcon/TokenIconWithGIVBack';
@@ -41,6 +43,7 @@ import StorageLabel from '@/lib/localStorage';
 import InlineToast, { EToastType } from '@/components/toasts/InlineToast';
 import { useAppSelector } from '@/features/hooks';
 import { useModalCallback } from '@/hooks/useModalCallback';
+import { useGeneralWallet } from '@/providers/generalWalletProvider';
 import EligibilityBadges from '@/components/views/donate/common/EligibilityBadges';
 import EstimatedMatchingToast from '../../EstimatedMatchingToast';
 
@@ -70,6 +73,7 @@ export const QRDonationCard: FC<QRDonationCardProps> = ({
 	const { modalCallback: signInThenDonate } = useModalCallback(() =>
 		setShowDonateModal(true),
 	);
+	const { isConnected } = useGeneralWallet();
 
 	const {
 		project,
@@ -90,7 +94,7 @@ export const QRDonationCard: FC<QRDonationCardProps> = ({
 		retrieveDraftDonation,
 	} = useQRCodeDonation(project);
 
-	const { addresses, id } = project;
+	const { addresses, id, isGivbackEligible } = project;
 	const draftDonationId = Number(router.query.draft_donation!);
 	const [amount, setAmount] = useState(0n);
 	const [usdAmount, setUsdAmount] = useState(0);
@@ -106,6 +110,21 @@ export const QRDonationCard: FC<QRDonationCardProps> = ({
 	const isOnEligibleNetworks = selectedQFRound?.eligibleNetworks?.includes(
 		config.STELLAR_NETWORK_NUMBER,
 	);
+	const isProjectGivbacksEligible = !!isGivbackEligible;
+	const isTokenGivbacksEligible = !!stellarToken?.isGivbackEligible;
+	const isGivbacksEligible =
+		isProjectGivbacksEligible && isTokenGivbacksEligible;
+
+	// Stellar QR donations are matched without connecting a wallet — QF matching
+	// eligibility is surfaced by EligibilityBadges and the estimated matching UI.
+	// The only action worth prompting here is signing in for GIVbacks, and only
+	// when both the project and the selected token are GIVbacks eligible.
+	const showConnectWallet = isGivbacksEligible;
+
+	const textToDisplayOnConnect = () =>
+		isGivbacksEligible
+			? 'label.sign_into_giveth_for_a_chance_to_win_givbacks'
+			: null;
 
 	const donationUsdValue =
 		(tokenPrice || 0) * Number(ethers.utils.formatEther(amount));
@@ -313,6 +332,18 @@ export const QRDonationCard: FC<QRDonationCardProps> = ({
 					})}
 				/>
 			)}
+			{!showQRCode &&
+				!isConnected &&
+				!showEstimatedMatching &&
+				showConnectWallet &&
+				textToDisplayOnConnect() && (
+					<ConnectWallet>
+						<IconWalletOutline24 color={neutralColors.gray[700]} />
+						{formatMessage({
+							id: textToDisplayOnConnect() || '',
+						})}
+					</ConnectWallet>
+				)}
 			{!showQRCode && (
 				<EligibilityBadges
 					amount={amount}
@@ -426,6 +457,10 @@ export const QRDonationCard: FC<QRDonationCardProps> = ({
 		</>
 	);
 };
+
+const ConnectWallet = styled(BadgesBase)`
+	margin-bottom: 5px;
+`;
 
 const CardHead = styled(Flex)`
 	align-items: center;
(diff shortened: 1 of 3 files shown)


src/components/views/donate/common/EligibilityBadges.tsx
```diff
@@ -79,9 +83,8 @@ const EligibilityBadges: FC<IEligibilityBadges> = props => {
 				? 'page.donate.unlocks_matching_funds'
 				: null; // Prevents invalid id values
 
-	// Stellar (QR) donations don't require a connected wallet, so always show
-	// the eligibility badges for the Stellar flow — same as a connected wallet.
-	return isConnected || isStellar ? (
+	return isConnected ||
+		(isStellar && isStellarOnlyRound && !isProjectGivbacksEligible) ? (
```
RamRamez: **1. Coverage gap: project eligible but token not eligible → nothing renders.**

For a disconnected Stellar user, badges now show only when `!isProjectGivbacksEligible`, while the new prompt in `QRDonationCard` shows only when `isProjectGivbacksEligible && stellarToken?.isGivbackEligible`. If the project is eligible but the Stellar token is **not** (this component explicitly handles `page.donate.token_not_givbacks_eligible`), the user sees *neither* the token warning *nor* the prompt. The two conditions gate on different facts (badges: project only; prompt: project AND token), so they don't partition the state space.

RamRamez (commented):
## Code Review

Small, focused change and the intent is sound, but the two new visibility conditions don't compose cleanly — there are disconnected-user states where neither the badges nor the new sign-in prompt render, and the badge gating uses round shape as a proxy in a way that produces inconsistent behavior across rounds. Details in the inline comments (numbered by priority).

**What looks good:**
- The translation key `label.sign_into_giveth_for_a_chance_to_win_givbacks` exists in all locale files (en/es/ct).
- Reusing `BadgesBase` keeps the prompt visually consistent with the badges it replaces; the optional `warning`/`active` props default safely.
- The explanatory comment about why Stellar QR donations don't need a connected wallet is helpful context.

**Minor (no inline comment):** in `QRDonationCard.tsx`, the new local `isGivbacksEligible` (project AND token) differs by one character from the project field `isGivbackEligible`, and `EligibilityBadges.tsx` already has its own different `isGivbacksEligible` (which additionally requires the USD threshold). Something like `showGivbacksSignInPrompt` would prevent misreads.

src/components/views/donate/common/EligibilityBadges.tsx
```diff
@@ -66,6 +66,10 @@ const EligibilityBadges: FC<IEligibilityBadges> = props => {
 		isProjectGivbacksEligible &&
 		donationUsdValue >= GIVBACKS_DONATION_QUALIFICATION_VALUE_USD;
 
+	const isStellarOnlyRound =
+		selectedQFRound?.eligibleNetworks?.length === 1 &&
+		selectedQFRound?.eligibleNetworks[0] === config.STELLAR_NETWORK_NUMBER;
```
RamRamez: **2. `isStellarOnlyRound` gating is inconsistent across round shapes.**

The "project not GIVbacks eligible" warning now shows to disconnected users only when the active QF round happens to contain exactly Stellar. The same disconnected user on the same non-eligible project sees the warning during a Stellar-only round, but nothing during a multi-network round or outside any round (`selectedQFRound` undefined → `isStellarOnlyRound` is false).

Project GIVbacks eligibility has nothing to do with round network composition, so this reads like the wrong predicate — if the goal is "warn about non-eligible projects in the QR flow," the round check shouldn't be here; if the goal is "hide everything for disconnected users," the `!isProjectGivbacksEligible` escape hatch shouldn't be. Which is intended?

src/components/views/donate/OneTime/SelectTokenModal/QRCodeDonation/QRDonationCard.tsx
```diff
@@ -313,6 +332,18 @@ export const QRDonationCard: FC<QRDonationCardProps> = ({
 					})}
 				/>
 			)}
+			{!showQRCode &&
+				!isConnected &&
```
RamRamez: **3. `isConnected` vs `isSignedIn` mismatch.**

The prompt text says "**Sign into** Giveth," but visibility is gated on wallet *connection* (`useGeneralWallet().isConnected`). A user who is connected but not signed in (the state `handleNext` explicitly handles with `isEnabled && !isSignedIn`) gets no prompt even though they still need to sign in for GIVbacks (`anonymous: isSignedIn && isEnabled ? false : true`). Gating on `isSignedIn` (or `!isEnabled || !isSignedIn`) would match the copy and the actual GIVbacks requirement.

src/components/views/donate/OneTime/SelectTokenModal/QRCodeDonation/QRDonationCard.tsx
```diff
@@ -313,6 +332,18 @@ export const QRDonationCard: FC<QRDonationCardProps> = ({
...
+				!showEstimatedMatching &&
+				showConnectWallet &&
+				textToDisplayOnConnect() && (
```
RamRamez: **4. Dead/redundant logic in the prompt condition.**

`showConnectWallet` and `textToDisplayOnConnect()` are both derived solely from `isGivbacksEligible`, so this guard checks the same boolean three times, and the `|| ''` fallback inside `formatMessage` below is unreachable. It collapses to:

```tsx
{!showQRCode && !isConnected && !showEstimatedMatching && isGivbacksEligible && (
    <ConnectWallet>
        <IconWalletOutline24 color={neutralColors.gray[700]} />
        {formatMessage({ id: 'label.sign_into_giveth_for_a_chance_to_win_givbacks' })}
    </ConnectWallet>
)}
```

src/components/views/donate/OneTime/SelectTokenModal/QRCodeDonation/QRDonationCard.tsx
```diff
@@ -313,6 +332,18 @@ export const QRDonationCard: FC<QRDonationCardProps> = ({
 					})}
 				/>
 			)}
+			{!showQRCode &&
+				!isConnected &&
+				!showEstimatedMatching &&
```
RamRamez: **5. Prompt disappears once the donation qualifies for matching.**

The `!showEstimatedMatching` guard means that as soon as the user enters an amount above the matching minimum, the sign-in prompt vanishes — but signing in is still required to receive GIVbacks at that point. If this is to avoid visual clutter, consider whether the prompt should instead coexist with (or move below) the matching toast.

**6. Prompt isn't actionable.** It tells the user to sign in but isn't clickable — `signInThenDonate` / the sign-in modal callback is already wired up in this component, so an `onClick` (with cursor styling) would make the call-to-action functional.

RamRamez (approved):
Thanks @kkatusic

src/components/views/donate/common/EligibilityBadges.tsx
kkatusic: resolved

src/components/views/donate/common/EligibilityBadges.tsx
kkatusic: resolved

src/components/views/donate/OneTime/SelectTokenModal/QRCodeDonation/QRDonationCard.tsx
kkatusic: resolved

src/components/views/donate/OneTime/SelectTokenModal/QRCodeDonation/QRDonationCard.tsx
kkatusic: resolved

